PR #1, opened 2016-07-21
Any type

nadako:
This proposes introduction of the new `Any` type as a safer alternative to `Dynamic` when it comes to simply holding "any value".

[Rendered version here](https://github.com/nadako/haxe-evolution/blob/any/proposals/0000-any.md)

### DIFF
--- a/proposals/0000-any.md
+++ b/proposals/0000-any.md
@@ -129,6 +129,10 @@ I was considering stripping down `Dynamic` to have the same meaning as the propo
  * The name `Dynamic` is long and suggests some kind of dynamic behaviour, so it's better suited for the current
    version of `Dynamic`.
 
+## Opening possibilities
+
+If we embrace the `Any` type and leave `Dynamic` for the "really dirty low-level stuff", we could not take `Dynamic` into consideration when designing and implementing new type system features, such as method overloading, which is surely going to make life easier.
+
 ## Unresolved questions
 
  * Whether or not the `Any` abstract type should have direct cast `to Dynamic`, see the design section for more info.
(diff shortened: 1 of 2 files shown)


ncannasse:
Is there any change in standard library where it would make sense to use Any instead of Dynamic?

nadako:
As mentioned in the proposal, `haxe.Unserializer.run` and `haxe.Json.parse` are good candidates for this.

I would also use `Any` instead of `Dynamic` for method signatures of `Reflect`, since IMO reflection is a higher level concept than dynamic, and staying type-safe there would be a good thing.

But changing any of this would be a breaking change, so I'm not proposing to do it for now. Maybe we can revise that for Haxe 4.

nadako:
Ping!

nadako:
Unfortunately, the discussion took place in the internal Slack chat instead of this PR, but here's the summary:
 * in general: the proposal is good
 * about unresolved question of autocasting: it should autocast
 * about standard library usage: add Any usage to unserializers and reflection, but hide it behind a define, so by default it uses Dynamic like before, but can be easily switched to Any, so people can test their codebases
 * about when to add this: should be safe for 3.3 release (both Any and #if's)

I'll now adjust the proposal text and merge it. After that I'm gonna adjust and merge the [implementation PR](https://github.com/HaxeFoundation/haxe/pull/5439) to Haxe repository.

ncannasse:
Please note that we didn't correctly respected the protocol here : have a consensus between me and Dan does not mean we have a majority to accept the HXP.

nadako:
You're right... That was unfortunate. I was too excited to get it merged... We need to get everyone to vote somehow...

PR #2, opened 2016-08-01
Formalize the process

ncannasse:


nadako:
as discussed, this needs some wording fixes, but everyone is fine with the actual content, so i'm merging this

PR #7, opened 2016-10-11
Spread `...` operator

zabojad:
Hi Haxe team !

I'd like to propose a ES6 like `...` spread & rest operator.

Hope we will have a constructive discussion about it.

Best regards,

Thomas

[Rendered version](https://github.com/zabojad/haxe-evolution/blob/master/proposals/0002-spread-operator.md)


proposals/0002-rest-spread-operator.md
```diff
@@ -0,0 +1,133 @@
+# Feature name
```
nadako: This one is supposed to be changed, I'll make it more obvious in the template :)

proposals/0002-rest-spread-operator.md
```diff
@@ -0,0 +1,133 @@
...
+  
+  This:
+  ```
+  function pure(data : MyStruct) {
+
+  	return {...data, overwrittenFieldName: overwrittenFieldValue};
+  }
+  ```
+
+  is much less verbose than this:
```
nadako: It's not just verbose, I think it's also worse performance, wise. If for `...obj, field2: somevalue` we generate an object decl like `{field1: obj.field1, field2: somevalue}`, the output will be cleaner and more performant.

proposals/0002-rest-spread-operator.md
```diff
@@ -0,0 +1,133 @@
...
+
+  	ret.overwrittenFieldName = overwrittenFieldValue;
+
+  	return ret;
+  }
+  ```
+
+* The standard javascript externs do not expose the [Object.assign API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign).
+
+  While it's not complicated to write an extern class for this, it feels wrong to add it to the source code of a custom project while it is something that should be available by default.
```
nadako: Speaking of which, we should probably have modern JS standard APIs in our std lib, but that's outside of this proposal's scope.

proposals/0002-rest-spread-operator.md
```diff
@@ -0,0 +1,133 @@
...
+
+* This `...` operator is part of the ECMAScript standard which Haxe syntax follows to some point. Getting closer to the ECMAScript syntax would ease the adoption of Haxe by javascript developers. 
+
+* Also, if added to the Haxe syntax, we would also get several other benefits out of it:
+```
+// array decomposition (non existent in Haxe)
+myFunction(...myArray);
```
nadako: I don't think this is useful in Haxe, since our Arrays are typed and arguments tend to have different types. But if we'll have tuples someday, this syntax could be used for unpacking them into arguments.

proposals/0002-rest-spread-operator.md
```diff
@@ -0,0 +1,133 @@
...
+// shorter and more powerful array literals / comprehension
+var myNewArr = [ 1, 4, ...myOldArr, 56 ];
```
nadako: Not sure how useful it is to be honest. Also this will obviously require looping, but OTOH I'm not against it, since it looks pretty clear what's happening here.

proposals/0002-rest-spread-operator.md
```diff
@@ -0,0 +1,133 @@
...
+// even augmented push
+myOldArr.push(...myNewArr);
```
nadako: Not gonna happen, Haxe's Array.push doesn't work like that.

proposals/0002-rest-spread-operator.md
```diff
@@ -0,0 +1,133 @@
...
+// object duplication
+var myNewObj = { ...myOldObj }
```
nadako: This (and the second one) are very useful IMO. This is similar to OCaml's `{obj with field = value}` which I find great for persistent data structures.

proposals/0002-rest-spread-operator.md
```diff
@@ -0,0 +1,133 @@
...
+
+
+## Detailed design
+
+There are two cases to differentiate when using the `...` operator:
+
+* [On Objects](https://babeljs.io/docs/plugins/transform-object-rest-spread/):
+```
+// Spread properties
+var n = { x, y, ...z };
```
nadako: This example assumes that Haxe has the `{x, y}` syntax, which it doesn't, and it's not related to the spread operator proposal, please remove.

proposals/0002-rest-spread-operator.md
```diff
@@ -0,0 +1,133 @@
...
+trace(n); // { x: 1, y: 2, a: 3, b: 4 }
+
+// Rest properties
+var { x, y, ...z } = { x: 1, y: 2, a: 3, b: 4 };
```
nadako: Same here, Haxe doesn't have object destructuring and it's a separate feature that was discussed and rejected before. I propose taking this step by step and restricting this proposal to one syntax for now: `{...obj, field: 1}` (well, maybe Arrays too, but I'm not sure).

proposals/0002-rest-spread-operator.md
```diff
@@ -0,0 +1,133 @@
...
+
+We gain a powerful object duplication operator that is missing in Haxe right now.
+
+
+* On Arrays:
+
+```
+// Spread elements
+var myArr = [ 1, 2, 3 ];
+myFunc(...myArr);  // here, would do the same as: myFunc(1,2,3);
```
nadako: These was described earlier, and it's quite clear what are they for. The "Detailed design" section is supposed to describe how it's going to work on the high level, for example:

* We need to introduce a new unary operator e.g. `OpSpread` that's parsed from `...<expr>` (which is actually a breaking change for macros, since adding new enum constructor will affect exhaustiveness checks).
* `{...obj, field2: 42}` will translate to `{field1: obj.field1, field2: 42}` (that's the idea, right?)
* `[1, 2, ...arr, 42, ...arr2, 1]` will translate to `var tmp = [1,2]; for (v in arr) tmp.push(v); tmp.push(42); for (v in arr2) tmp.push(v); tmp.push(1);`
* etc, etc.

RealyUniqueName:
Despite the name there is no section about rest arguments in this proposal )
Or is it a subject for another one?

nadako:
> Despite the name there is no section about rest arguments in this proposal )
Or is it a subject for another one?

I thought about this, but I see this as a completely different topic. Rest in Haxe is a bit weird thing, only used for externs, so I don't see any point in adding special syntax for it. But if we'll really want rest arguments at some point, I guess adding the `...args:String` or similar syntax makes sense.

proposals/0002-rest-spread-operator.md
zabojad: I think you wanted to highlight lines 57->61 here. Yes I agree with you, those would be the most useful additions.

zabojad:
Thanks for your review. I'm adding / correcting accordingly right now. I'll keep only the spread use case and remove the rest one.

I've initially requested both because it's how they are presented in ES6 docs / specifications.

zabojad:
OK, done.

The scope has been reduced to the spread operator only.

About the `detailed design` section, I'm not sure about what is possible and what is not so don't hesitate to make your own proposition there...

nadako:
So, here's my opinion:

I think it wouldn't hurt to have `OpSpread` (or whatever we call it) unary operator that can be both prefix and postfix (so we allow `expr...` and `...expr` syntax that becomes `EUnop(OpSpread, true/false, expr)`). This alone would allow some nice macro/abstract magic.

I can see use cases for the feature that translates `{...obj, field2: value}` to `{field1: obj.field1, field2: value}` and if it were there I'd probably use it. OTOH for me it's not critical or even "cool" feature and this is something totally doable with a simple macro function anyway. Also, it's not clear how this should be represented in the AST. I guess we could have a `{field: null, expr: EUnop(OpSpread, false, expr)}`, but I don't know, feels a bit weird... Also, I don't know how useful would it be to support merging multiple objects like that (`{...a, ...b, f: 10}`). It makes more sense in a dynamic language like JavaScript, but in Haxe, I'm not so sure.

It's not clear to me how to deal implementation-wise with object optional fields here. I guess we'd have to generate some `Reflect.hasField` calls which could generate some ugly code, but it's not a big deal.

Side note: if we ever implement this, it could probably even work well with the new `@:structInit` classes.

For Arrays (`[1, 2, ...otherArr, 3, ...anotherArr, 4]`), I feel like it's a "cool feature" that won't be used too much. One case I often need tho is `[...otherArr, "some"]` to make a new array with a new element. Basically similar to OCamls `"some" :: otherArr`. This is useful for persistent arrays. Right now, the easiest way to do that is `otherArr.concat(["some"])`, but that allocates a new array just for that element which sucks performance-wise.

Regarding `func(...argArray)`, I think it won't have uses at all, because, as I mentioned earlier, Haxe static nature just won't play well with this because 1) Array elements have just one type while function arguments are different most of the times; 2) We don't use Reflect.callMethod (AKA func.apply) much, because that has performance issues, I think even in JavaScript it's not recommended.

---

Regarding the proposal:

It's still a bit focused on JavaScript target, but with features like this we need to consider it for all targets, which means we should think of strict typing and static target performance. Also that JS `_extends` example looks horrible performance-wise. 😉 I get that it's the best you can do with JS, but I'm not sure it's a good example for Haxe.

The proposal says that this feature has no impact on existing code, but all additions like this will surely affect macro-related data structures which is a breaking change. And speaking of these structures, I think a syntax proposal should describe the new/changed AST nodes in the "detailed design" section.

Regarding "opening possibilities" section: I don't think this will help "marketing" within JS community at all, but OTOH this could be a nice feature for the growing functional programming community.

Also I doubt this is going to impact potential ES6 target, because that's just a Haxe syntax proposal, and that syntax is going to be typed and optimized by the compiler and on the generator level it probably lose the notion of original syntax at all. :)

zabojad:
@nadako ok I'll come back next week with a revised version of this proposal.

fullofcaffeine:
@zabojad Any updates?

zabojad:
@fullofcaffeine not yet

I'm a little bit overloaded with my current tasks so please let this request open and I'll complete it as soon as I can...

Simn:
I'm having a hard time seeing this as anything other than a shortcut for Reflect.callMethod and Array.concat. It strikes me as something that looks cool in small samples, but is never really needed to write expressive and readable code.

Furthermore, `...` already is an operator in Haxe.

Overall, I'm currently not in favor or adding this to the language.

fullofcaffeine:
> Furthermore, ... already is an operator in Haxe.

What operator is it? 

Could this be added as a macro then?

nadako:
It's an integer iterator literal `0...10`. I'm wondering tho if it would be possible to have it as an unary operator as well (e.g. `1...` or `...1`). Might be useful for macros/abstracts.

zabojad:
@Simn no problem, feel free to close this request then...

@nadako sorry I did not come back to you but what you ask (describing the new/changed AST nodes) is beyond my current knowledge and I do not really have time to look at it. Besides, I did not find any doc nor ressources on haxe.org or elsewhere that could help in doing that.

Just few words on the process : at first, I was quite excited by the possibility to propose new features to Haxe. However in practice, I think the process is a bit too demanding especially in this case : a spread operator is not something over complicated and already exists in other languages... Also I have the feeling that the team might not consider javascript a priority...

nadako:
> a spread operator is not something over complicated and already exists in other languages

Unfortunately that doesn't help us, because to implement a feature we still have to answer a lot if "what if" questions, which are supposed to be answered by the "detailed design" section. Also keep in mind that JavaScript is fully dynamic, while Haxe is not, so one can't just say "do whatever js does", OTOH unlike e.g. OCaml which has the similar feature, Haxe it's not fully static as well, because there's Dynamic, so we have to consider it and make _some_ decision regarding this.

zabojad:
@nadako I completely understand that you need to answer all the "what if" questions. But in my opinion, you should not expect the average Haxe user to be able to write alone the "detailed design" of your process without giving enough documentation on how the AST works, with perhaps few feature requests examples showing different cases covering what an average Haxe developper should know to be able to describe this section. 

I understand that with this process, you seek for quality and want to avoid regressions or bad design to be added to Haxe. But right now this process is too demanding if one should know exactly how the Haxe compiler works internally before submitting a request. If you want this process to be a bit more productive, you either assign the detailed design description to a Haxe team member or you provide the necessary ressources to every one so that we can actually do it.

zabojad:
@nadako Hi Dan, I'd like to move forward on this request. Do you have any ressource that could help me describe the `detailed design` section? Could give an hand on this?

nadako:
At this point I wouldn't put much effort into this unless @ncannasse approves the feature and its syntax in general.

ncannasse:
I think the feature can be already done with either macros or even normal function + reflection such as

```haxe
static function replace<F:{},O:F>( obj : O, fields : F ) : O {
    var obj = Reflect.copy(obj);
    for( f in Reflect.fields(fields) ) Reflect.setField(obj,f, Reflect.field(fields,f));
    return obj;
}
```

So I need to point in adding some specific syntax that requires extra learning.

pypmannetjies:
@ncannasse Can you give some more information on this suggested approach and usage of this function?

zabojad:
@ncannasse correct me if I'm wrong but what you suggest is to be used like this: `var myNewObj = replace(myOtherObj, { f1: "...", f2: "...", f3: "..." });`, is that right ? 

If that's the case, it misses the point in addressing the original issue + we already have this with `js.Object.assign()`.

The original point here was that in javascript today and with libraries like react, redux and more..., we very often need to copy and merge object(s) into new object(s). ES6 addresses it with a spread (and a rest) operator more or less as described in my original request. The point of the original request was not to find a way to copy objects or merge objects fields into another object, we perfectly know about `Reflect` of `js.Object`. The point of the original request was really about a new syntax that allows to do it in a much less verbose and more readable way thus the original request about a new syntax.

I don't think macros would allow us to write things like `var t = { ...a, add: "..." }`, am I right?

I know you do not like new syntaxes to be added to the language and I completely hear your points about it. But in my opinion, if Haxe targets js and allows the use of popular js libraries, we also need to keep the syntax practical for the uses it will be subjected to.

I proposed something similar to what ES6 introduced but it could be something else actually. Now it might be a more subjective opinion but I find this spread operator syntax pretty obvious and it has proven its capabilities in other ES6 already...

zabojad:
PS : maybe I'm completely wrong as I do not have a clue about how the compiler works but... With Haxe, couldn't we have even a sort of "spread operator on steroïds" :)? By this, I mean a spread operator that would for example print a warning at compilation when you spread fields into an object of a type that do not have those fields in its definition, or things like this...

ncannasse:
Right, my function was about replacing field values, not adding new ones.
To add fields and get type safety you need a macro. This does not require
new syntax.

zabojad:
@ncannasse I do agree that we can do all that with macros or functions and that doing this does not require a new syntax.

But what does require a new syntax? New features only? If yes, then my opinion diverges here. New syntax are also pertinent where repetitive and verbose pieces of code could be written otherwise shorter without loss of meaningful pieces of information.

pypmannetjies:
Agreed. This would be some nice sugar for the language.

nadako:
For the time being, I quickly hacked up a macro that allows for OCaml-like `with` functionality, which I believe the main use-case for the spread operator: https://gist.github.com/nadako/7edbc859e5206b2dbd65df2fd96aa334

It doesn't deal with optional fields yet, but it's something that's not hard to add. I'll prepare a Cookbook article later.

kevinresol:
You can also use [tink_anon](https://github.com/haxetink/tink_anon): `return tink.Anon.merge(player, level = player.level + 1)`

zabojad:
All these are very nice but way more verbose than a built-in spread operator...

Anyway, thank you all for your time spent on this...

nadako:
>  way more verbose than a built-in spread operator...

mhm, I don't know about that :-)
```haxe
player.with(level = player.level + 1)
{...player, level : player.level + 1}
```

(not that I'm against the syntax, but it's less about verbosity and more about clear meaning of the expression)

fullofcaffeine:
Nice macro Dan. Looking forward to reading the cookbook article :)

zabojad:
>>way more verbose than a built-in spread operator...
>mhm, I don't know about that :-)

Erf, OK I give you that one ^^.

Now how would your macro perform in terms of compilation performances and impact on auto-completion vs a built-in syntax :)?

nadako:
> Now how would your macro perform in terms of compilation performances and impact on auto-completion vs a built-in syntax :)?

Surely it'll be slower than something implemented directly in the compiler, but I think it's not that bad, since it doesn't do much compiler<->macro data conversion. It does the typing twice atm though, but that can be improved using `Context.storeTypedExpr`.

As for the completion - we don't have good completion for anon structure fields right now anyway, so there's not much difference :) If we work on HaxeFoundation/haxe#6786, we can improve autocompletion for macro-processed expressions (in this case - provide toplevel completion for fields inside the `with(...)` call.

zabojad:
@nadako OK OK, I surrender :)... I'm really looking forward to reading your Cookbook article about it! Thank you again for your efforts!

vonagam:
If the problem is that the spread operator requires new syntax that does not yet present in haxe in any form, then those two variants can be used, mirroring haxe's anon structures type notations:
```haxe
var next = {>prev, key: value}; // as extension notation
var next = prev & {key: value}; // as intersection notation
```
Second variant can be implemented without even touching the parser.

PR #8, opened 2016-10-11
Arrow functions

RealyUniqueName:
This proposes better syntax for anonymous functions for the sake of cleaner and easier to maintan code

[Rendered version](https://github.com/RealyUniqueName/haxe-evolution/blob/master/proposals/0000-arrow-functions.md)

### DIFF
--- a/proposals/0000-arrow-functions.md
+++ b/proposals/0000-arrow-functions.md
@@ -103,14 +103,14 @@ Following syntax is proposed for arrow functions in Haxe:
     function(arg1, arg2) return expr;
     ```
 
-* Implicit typing:
+* Explicit typing:
 
     ```haxe
     (arg1:Int, arg2:String) -> expr
     //equivalent for
     function(arg1:Int, arg2:String) return expr;
     ```
-    Implicit return type for arrow functions should not be allowed to not interfer with existing syntax of function type.
+    Explicit return type for arrow functions should not be allowed to not interfer with existing syntax of function type.
 
 ### AST
 
--- a/proposals/0000-arrow-functions.md
+++ b/proposals/0000-arrow-functions.md
@@ -5,7 +5,7 @@
 
 ## Introduction
 
-Provide better syntax for anonymous functions declaration. 
+Provide better syntax for anonymous functions declaration.
 
 ## Motivation
 
@@ -14,10 +14,10 @@ Provide better syntax for anonymous functions declaration.
     ```haxe
     //hard to read
     var names = users.map(function(u) return u.getProfile()).filter(function(p) return p.inCountry('USA')).map(function(p) return p.getName());
-    
+
     //a little better
     var profiles = users.map(function(u) return u.getProfile());
-    profiles = profiles.filter(function(p) return p.inCountry('USA'));    
+    profiles = profiles.filter(function(p) return p.inCountry('USA'));
     var names = profiles.map(function(p) return p.getName());
 
     //best
@@ -30,18 +30,18 @@ Provide better syntax for anonymous functions declaration.
     //vs
     array.map(a -> a.toInt()).sort((a, b) -> a - b);
     ```
-    Required boilerplate reduced from 32 chars to 4. 
+    Required boilerplate reduced from 32 chars to 4.
 
-* Nowadays arrow functions are largely adopted by programmers community. 
+* Nowadays arrow functions are largely adopted by programmers community.
     Lack of such functions in Haxe raises questions for newcomers that Haxe suffers stagnation in development.
-    This factor has negative impact on growth of Haxe community.  
+    This factor has negative impact on growth of Haxe community.
 
 * Current syntax of anonymous functions in Haxe is quite verbose compared to other languages.
     ```haxe
     //Haxe
     function() return expr
     function(arg) return expr
-    
+
     //TypeScript
     () => expr
     arg => expr
@@ -57,11 +57,11 @@ Provide better syntax for anonymous functions declaration.
     //Scala
     () => expr
     (arg) => expr
-        
+
     //Java
     () -> expr
     (arg) -> expr
-    
+
     //Swift
     { expr }
     { arg1 in expr }
@@ -75,9 +75,9 @@ Provide better syntax for anonymous functions declaration.
 
 ## Detailed design
 
-### Syntax 
+### Syntax
 
-Following syntax is proposed for arrow functions in Haxe: 
+Following syntax is proposed for arrow functions in Haxe:
 
 * No arguments:
 
@@ -87,7 +87,7 @@ Following syntax is proposed for arrow functions in Haxe:
     function() expr; //for Void->Void
     function() return expr;
     ```
-    
+
 * Single argument:
 
     ```haxe
@@ -109,19 +109,22 @@ Following syntax is proposed for arrow functions in Haxe:
 * Implicit typing:
 
     ```haxe
-    (arg1:Int, arg2:String):Int -> expr
+    (arg1:Int, arg2:String) -> expr
     //equivalent for
-    function(arg1:Int, arg2:String):Int return expr;
+    function(arg1:Int, arg2:String) return expr;
+    function(arg1:Int, arg2:String) expr;
     ```
+    Implicit return type for arrow functions should not be allowed to not interfer with existing syntax of function type.
 
-If return type is not specified, then whether function returns something or has `Void` return type, should be decided by type inference system. 
+    Whether function returns something or has `Void` return type, should be decided by type inference system.
 
 ### AST
 
 Following constructor is proposed for expression definition of arrow functions:
 ```haxe
-ELambda(args:Array<FunctionArg>, ret:Null<ComplexType>, expr:Expr)
+ELambda(args:Array<FunctionArg>, expr:Expr)
 ```
+Notice an abscense of return type.
 
 ## Impact on existing code
 
@@ -137,4 +140,4 @@ Since macros are being executed before typing, it can be hard to find out in mac
 
 Since users got used to use arrow functions in other languages, various implementations were created with macros: at least two librarys on haxelib
 and few more projects in outher sources.
-However such implementations impact compilation time, introduce non-standart (and different) syntax and should be avoided inside of other macros (breaks compiler cache)  
+However such implementations impact compilation time, introduce non-standart (and different) syntax and should be avoided inside of other macros (breaks compiler cache)
--- a/proposals/0000-arrow-functions.md
+++ b/proposals/0000-arrow-functions.md
@@ -1,7 +1,7 @@
 # Arrow functions
 
 * Proposal: [HXP-0000](0000-arrow-functions.md)
-* Author: [Haxe Developer](https://github.com/RealyUniqueName)
+* Author: [Alexander Kuzmenko](https://github.com/RealyUniqueName)
 
 ## Introduction
 
--- a/proposals/0000-arrow-functions.md
+++ b/proposals/0000-arrow-functions.md
@@ -107,9 +107,10 @@ Following syntax is proposed for arrow functions in Haxe:
 * Implicit typing:
 
     ```haxe
-    (arg1:Int, arg2:String):Void => expr
+    (arg1:Int, arg2:String):Int => expr
     //equivalent for
-    function(arg1:Int, arg2:String):Void expr;
+    function(arg1:Int, arg2:String):Int return expr;
+    
     ```
 
 * Add brackets if `=>` operator can interfer map declaration:
--- a/proposals/0000-arrow-functions.md
+++ b/proposals/0000-arrow-functions.md
@@ -28,7 +28,7 @@ Provide better syntax for anonymous functions declaration.
     ```haxe
     array.map(function(a) return a.toInt()).sort(function(a, b) return a - b);
     //vs
-    array.map(a -> a.toInt()).sort((a, b) => a - b);
+    array.map(a -> a.toInt()).sort((a, b) -> a - b);
     ```
     Required boilerplate reduced from 32 chars to 4. 
 
--- a/proposals/0000-arrow-functions.md
+++ b/proposals/0000-arrow-functions.md
@@ -75,6 +75,8 @@ Provide better syntax for anonymous functions declaration.
 
 ## Detailed design
 
+### Syntax 
+
 Following syntax is proposed for arrow functions in Haxe: 
 
 * No arguments:
@@ -122,6 +124,13 @@ Following syntax is proposed for arrow functions in Haxe:
     ```
 
 If return type is not specified, then whether function returns something or has `Void` return type, should be decided by type inference system. 
+
+### AST
+
+Following constructor is proposed for expression definition of arrow functions:
+```haxe
+ELambda(args:Array<FunctionArg>, ret:Null<ComplexType>, expr:Expr)
+```
 
 ## Impact on existing code
 
--- a/proposals/0000-arrow-functions.md
+++ b/proposals/0000-arrow-functions.md
@@ -84,7 +84,6 @@ Following syntax is proposed for arrow functions in Haxe:
     ```haxe
     () -> expr
     //equivalent for (depending on context)
-    function() expr; //for Void->Void
     function() return expr;
     ```
 
@@ -93,7 +92,6 @@ Following syntax is proposed for arrow functions in Haxe:
     ```haxe
     arg -> expr
     //equivalent for
-    function(arg) expr;
     function(arg) return expr;
     ```
 
@@ -102,7 +100,6 @@ Following syntax is proposed for arrow functions in Haxe:
     ```haxe
     (arg1, arg2) -> expr
     //equivalent for
-    function(arg1, arg2) expr;
     function(arg1, arg2) return expr;
     ```
 
@@ -112,11 +109,8 @@ Following syntax is proposed for arrow functions in Haxe:
     (arg1:Int, arg2:String) -> expr
     //equivalent for
     function(arg1:Int, arg2:String) return expr;
-    function(arg1:Int, arg2:String) expr;
     ```
     Implicit return type for arrow functions should not be allowed to not interfer with existing syntax of function type.
-
-    Whether function returns something or has `Void` return type, should be decided by type inference system.
 
 ### AST
 
--- a/proposals/0000-arrow-functions.md
+++ b/proposals/0000-arrow-functions.md
@@ -9,7 +9,7 @@ Provide better syntax for anonymous functions declaration.
 
 ## Motivation
 
-* Even simple closures encourage splitting expressions into multiple expressions, which makes code too bloated.
+* Even simple closures encourage splitting expressions into multiple lines, which makes code too bloated.
     It's hard to read and maintain code that heavily relies on anonymous functions.
     ```haxe
     //hard to read
--- a/proposals/0000-arrow-functions.md
+++ b/proposals/0000-arrow-functions.md
@@ -114,11 +114,7 @@ Following syntax is proposed for arrow functions in Haxe:
 
 ### AST
 
-Following constructor is proposed for expression definition of arrow functions:
-```haxe
-ELambda(args:Array<FunctionArg>, expr:Expr)
-```
-Notice an absence of return type.
+To distinct normal and arrow function syntaxes in AST new field `@:optional isLambda:Bool` is proposed to be added to [haxe.macro.Function](http://api.haxe.org/haxe/macro/Function.html) structure. Such change will have minimal impact on existing macros.
 
 ## Impact on existing code
 
--- a/proposals/0000-arrow-functions.md
+++ b/proposals/0000-arrow-functions.md
@@ -9,7 +9,7 @@ Provide better syntax for anonymous functions declaration.
 
 ## Motivation
 
-* Even simple closures encourages splitting expressions into multiple expressions, which makes code too bloated.
+* Even simple closures encourage splitting expressions into multiple expressions, which makes code too bloated.
     It's hard to read and maintain code that heavily relies on anonymous functions.
     ```haxe
     //hard to read
@@ -18,7 +18,7 @@ Provide better syntax for anonymous functions declaration.
     //a little better
     var profiles = users.map(function(u) return u.getProfile());
     profiles = profiles.filter(function(p) return p.inCountry('USA'));    
-    var names = profile.map(function(p) return p.getName());
+    var names = profiles.map(function(p) return p.getName());
 
     //best
     var names = users.map(u => u.getProfile()).filter(p => p.inCountry('USA')).map(p => p.getName());
@@ -30,9 +30,9 @@ Provide better syntax for anonymous functions declaration.
     //vs
     array.map(a => a.toInt()).sort((a, b) => a - b);
     ```
-    Required boilerplate reduced from 28 chars to 6. 
+    Required boilerplate reduced from 32 chars to 4. 
 
-* Nowadays arrow functions largely adopted by programmers community. 
+* Nowadays arrow functions are largely adopted by programmers community. 
     Lack of such functions in Haxe raises questions for newcomers that Haxe suffers stagnation in development.
     This factor has negative impact on growth of Haxe community.  
 
@@ -57,14 +57,14 @@ Provide better syntax for anonymous functions declaration.
     //Scala
     () => expr
     (arg) => expr
-
-    //Swift
-    { expr }
-    { arg1 in expr }
-
+        
     //Java
     () -> expr
     (arg) -> expr
+    
+    //Swift
+    { expr }
+    { arg1 in expr }
 
     //Ruby
     ->() {expr}
@@ -104,7 +104,7 @@ Following syntax is proposed for arrow functions in Haxe:
     function(arg1, arg2) return expr;
     ```
 
-* Multiple exprs in function body:
+* Multiple expressions in function body:
 
     ```haxe
     (arg1, arg2) => { exprs }
--- a/proposals/0000-arrow-functions.md
+++ b/proposals/0000-arrow-functions.md
@@ -114,7 +114,9 @@ Following syntax is proposed for arrow functions in Haxe:
 
 ### AST
 
-To distinct normal and arrow function syntaxes in AST new field `@:optional isLambda:Bool` is proposed to be added to [haxe.macro.Function](http://api.haxe.org/haxe/macro/Function.html) structure. Such change will have minimal impact on existing macros.
+Since proposed syntax is just another way to express function declaration, no AST changes required.
+
+But in Haxe 4.0 new field `@:optional isLambda:Bool` could be added to [haxe.macro.Function](http://api.haxe.org/haxe/macro/Function.html) structure to distinct normal and arrow function syntaxes in AST.
 
 ## Impact on existing code
 
--- a/proposals/0000-arrow-functions.md
+++ b/proposals/0000-arrow-functions.md
@@ -37,7 +37,7 @@ Provide better syntax for anonymous functions declaration.
     This factor has negative impact on growth of Haxe community.  
 
 * Current syntax of anonymous functions in Haxe is quite verbose compared to other languages.
-    ```
+    ```haxe
     //Haxe
     function() return expr
     function(arg) return expr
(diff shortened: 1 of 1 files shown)


ncannasse:
Let's examine what would be the impact on the parser. 

For instance the following `(expr : Type)` is already parsed as `ECheckType(expr,type)`

We already have an `=>` operator, but we don't have a way to parse "tuples" such as `(a,b,c)` or with optional types `(a:Int,b,c:String)`

So this would require to change `ECheckType of expr * ttype` into `ETuple of (expr * ttype option) list` (maybe keep ECheckType for backward compatibility with a single case).

While Map literals and comprehensions can be type inferred to distinguish them from a arrow function, this is not very elegant since you can get a whole different meaning from `[x => "Hello"]` depending on the type it's inferred which is weird. The same goes for `[for( x in strings ) x => x.length]` which can have both a meaning as creating an Array of functions or a `Map<String,Int>`

I guess that would require to change Map literals & comprehensions syntax as well.

nadako:
A good point about map/array literal ambiguity, but I don't really understand how you want to represent this in untyped AST? `EBinop(OpArrow, ETuple([(e1, t1); (e2, t2)]), ef)`? Really? I think this should be parsed into a proper AST node, either `EFunction` or a new `EArrowFunction` (similar to how we have `EIf` and `ETernary`). Otherwise it'll be a huge pain working with them in macros.

clemos:
It's probably too many breaking changes, but I think it would make (or have made) sense to use `->` for maps (like in Scala) and `=>` for arrow functions (ES6, Scala).
But then again maybe it brings even more ambiguous syntaxes with existing `->` for function types, etc.

RealyUniqueName:
> While Map literals and comprehensions can be type inferred to distinguish them from a arrow function, this is not very elegant since you can get a whole different meaning from [x => "Hello"] depending on the type it's inferred which is weird. The same goes for [for( x in strings ) x => x.length] which can have both a meaning as creating an Array of functions or a Map<String,Int>

In such cases  user has to enclose expression in parenthesis, if he wants it to be interpreted as a closure einstead of a map.
```
[(x => "hello")]; 
[for( x in strings ) (x => x.length)]
```

RealyUniqueName:
I think `[x => "Hello"]` and `[for( x in strings ) x => x.length]`  should always be a map regardless expected type. Otherwise it's confusing if sometimes it's a map, but sometimes - an array of functions.

But with parentheses both cases become obvious.

RealyUniqueName:
Also we can consider using `->` instead of `=>`

ncannasse:
It's pretty confusing to either have parenthesis affecting the type of the expression. And I'm not in favor of having both `->` and `=>` since Haxe is not Perl :)

clemos:
> And I'm not in favor of having both `->` and `=>` Haxe is not Perl :) 

Or is it ;) ? 
https://github.com/HaxeFoundation/haxe/blob/development/src/syntax/lexer.mll#L287-L289

RealyUniqueName:
`->` is also used by Java, Ruby and Ocaml. But i still prefer `=>` )

RealyUniqueName:
> It's pretty confusing to either have parenthesis affecting the type of the expression

We already have this problem with abstracts operator overloading:
```
b + (a + c)
``` 
Is not guaranteed to be the same type as 
```
(b + a) + c
```
if those vars are abstracts.

back2dos:
I think by "type of expression" Nicolas did not mean types in the sense of our type system, but rather whether parentheses make the difference between a function literal or a binary operation.

To expand on that `=>` is a valid binary operator and abstracts may define it: http://try.haxe.org/#c3A05 - Changing this is a pretty big deal.

If we want to have something any time soon, then `->` is the way to go. I fail to see how distinguishing function syntax from a binary operator by using a different symbol is Perlish. I'd say relying on type is far crazier. Just use `->` instead. It's explicit. It's aligned with quite a few other languages.

ncannasse:
@RealyUniqueName that's not very common, and also unless you have very strange abstract types the semantics of + will most likely be preserved

RealyUniqueName:
Perhaps using `->` has more sense than i thought. This operator is already tied to functions in Haxe :)

nadako:
Yes, `->` could be good solution (tho it feels a bit uncomfortable after c# and js, but I guess it's kind of personal), but speaking of function type syntax, is everyone okay with it? Because if not, that's something that could be discussed too and changing it could impact arrow choice here :)

back2dos:
As far as function type syntax goes, my issue is rather that you can't specify parameter names.

For everything else, I don't care. It's an arrow. That's good enough. If there's any "correct" way to do it, it's to reuse mathematical notation for functions, in which case `ℝ ⟶ ℝ` is the type and `x ⟼ 2x` is the function itself. Make of that what you will wish ^^

RealyUniqueName:
> speaking of function type syntax, is everyone okay with it? 

It lacks arguments names. But i think with args names it may become too verbose.

Also `Int->String->Float != (Int->String)->Float` is a bit strange (again to parentheses affecting type question ;) )

RealyUniqueName:
I bump this thread.

Added [AST section](https://github.com/RealyUniqueName/haxe-evolution/blob/master/proposals/0000-arrow-functions.md#ast)

(summon @Simn, @andyli, @hughsando, @waneck)

hughsando:
I'm not too concerned about the syntax, but I would question whether it should be differentiated from the equivalent function definition in the AST. ie, after parsing, should you be able to tell the difference.
And if you can, might it be a variation of the exiting enum?

Simn:
Note that currently `arg => expr` is parsed as `EBinop(OpArrow,arg,expr)` and I don't think we want to change this.

I feel like `=>` is gonna be annoying to get right. There's a syntactic ambiguity because `[a => b]` could be either a map declaration or an array with an arrow function.

RealyUniqueName:
Are you ok with `->` instead of `=>`?

RealyUniqueName:
> I'm not too concerned about the syntax, but I would question whether it should be differentiated from the equivalent function definition in the AST

I think this case is somewhat similar to `EIf/ETernary`

Simn:
> Are you ok with -> instead of =>?

Yes, that would be my preferred solution anyway.

ncannasse:
I don't think that having both `->` and `=>` at the expression level is viable.
I know we already have both but atm one is for types and the other for expressions, so there's not much room for mistake.

Especially if other languages have been adopting `=>` for the same thing already, the amount of people that will use it instead of `->` by accident and turn into a misleading error message seems quite problematic.

markknol:
Summing up some issues

`=>` can be problematic with Map syntax. since current syntax allows this:
```haxe
[for(i in 0...10) i => 1]; // Map<Int, Int>
[for(i in 0...10) (i => 1)]; // Map<Int, Int>
[for(i in 0...10) (i) => 1]; // Map<Int, Int>
```

`->` can be problematic when you define a return type that contains function. Possible 
```haxe
(a:Type,b:Type):ReturnType -> returnValue 
// how?
(a:Type,b:Type):Void-Void -> returnValue 
(a:Type,b:Type):(Void-Void) -> returnValue 
```
We can choose to disallow return types and only allow this
```
(a:Type,b:Type) -> (returnValue:ReturnType)
(a:Type,b:Type) -> (returnValue:Void-Void)
```

frabbit:
one argument for using `->` is that we use the same syntax for function types, so the analogy is already there. Scala for example uses `=>`, but for both, function types and arrow functions.

frabbit:
I would argue that explicitly specifying the return type for arrow functions is not very important, because in almost 99% of all use cases for a short syntax it's not neccessary. And In such cases you can use `a -> (x:Int)` or just fallback to regular function syntax.

ousado:
I read through the comments, and it seems almost everyone is okay with or even in favor of `->` - is that correct?

What's best/better in terms of backwards compatibility?
`->` doesn't occur in expression positions so far, that's a clear winner.

What's the impact on the compiler? Almost zero, see my [implementation of the proposal here.](https://github.com/ousado/haxe/commit/1485a3c7b716eb024fd44c176c9696c4e12b4bca) 
I don't think it gets less intrusive than that.

What should be more important, a sense of familiarity for potential X,Y and Z users or the immediate benefit and complete backwards compatibility for existing Haxe users? I see exactly one possible answer, for anyone who's taking Haxe as a language seriously. 

Is "`=>` was taken already, so we used `->`" not a far better argument for the choice to everyone, including newcomers from X,Y and Z, than "we believed people familiar with X,Y and Z would like `=>` better, so we made our own users change their existing code and maintain two code bases for the transition period" - seriously, what kind of signal would that send?

So can we please just do this?

ousado:
Syntax as per proposal without return type before the arrow:
```haxe
class Test {

	static function main(){

		var a  = () -> 123;

		var b  =  a -> 123 * a;

		var b1 = (a) -> 123 * a;

		var b2 = (a:Int) -> 123 * a;

        var ff = a -> b -> b * a;

		var c =  a -> [for (i in 0...a) i * a];	
		
		var d = (a,b:Int) -> 123 * a + b;
		
		var e = (a:Int,b) -> 123 * a + b;

    }
}
```

RealyUniqueName:
Updated proposal with prohibition of implicit return type.

back2dos:
I take issue with this proposal (at least in its current form), because to me it lacks a clear goal beyond improving code golf handicap and following current hypes.

I do think Haxe could hugely benefit from additions of syntax that lends itself better for functional programming and having a concise notation for functions would seem to be the obvious place to start. This does not seem to be the motivation here, though:

1. Any function that returns `Void` is entirely about causing side effects and thus procedural.
2. Code written in functional style lends itself rather well for type inference. For that reason alone I think allowing to specify argument types is moot. Furthermore, if a function is so complex that it would benefit from type annotations, then it is generally a bad fit for a syntax that aims for terseness. Also, the "savings" in characters used become negligible.

Not that this has to be about functional programming at all. But it should be about something that goes further than shaving off a few characters. The addition of map literals and comprehensions was similarly ill-considered and look where it got us. Things should fit into some bigger picture - particularly in a programming language.

If it's hard to put a finger on what this is really about, then maybe we should bring some more real-worldish examples to the table, because the ones given so far fail to make a good point. Comprehensions provide more flexibility and better runtime characteristics (the equivalent code would allocate one array instead of three).

Given that the analyzer starts to understand purity, why not check short lambdas for such purity for example? Maybe that's a bad idea. Then let's have a better one. My point i, that after having argued about this for the better part of a decade, I think it would be a pity to then just hastily settle for a rather uninspired and aimless compromise.

Simn:
@back2dos: Can you give us a tl;dr on what you're actually suggesting?

fponticelli:
@back2dos, I don't agree with your point of view. I understand that we should aim to something more general but isn't this a valid starting point? What is wrong with this proposal that will make what you are looking for unobtainable?

Also I would say that for me personally, short lambdas really enhance readability a lot.

RealyUniqueName:
Changed [AST](https://github.com/RealyUniqueName/haxe-evolution/blob/master/proposals/0000-arrow-functions.md#ast) section to add a boolean field to existing function structure instead of introducing new expression definition.

skial:
I like @back2dos comment in point 2:

> Code written in functional style lends itself rather well for type inference. For that reason alone I think **allowing to specify argument types is moot**. Furthermore, **if a function is so complex that it would benefit from type annotations, then it is generally a bad fit for a syntax that aims for terseness**. Also, the "savings" in characters used become negligible

How about starting with a _restricted/limited_ version of what's being proposed, so if features need to be added, like explicit types, it should be easier to add than remove, especially if a significant amount of time has passed between a Haxe release.

Something along the lines of:

No args - `() -> expr`
Single args -  `(argN) -> expr`
Multiple args - `(arg, argN) -> expr`

And if explicit typing is needed in those extremely? rare situations, assigning to a variable should help, `var fn:Int->Int->Int = (x, y) -> x * y;` or using the `ECheckType`, `(((x, y) -> x * y):Int->Int->Int)`. The last one is horrible, but should work.

frabbit:
well the point is that haxe's type inference is, for different reasons, not as powerful as haskells for example. I think it's quite important to annotate types, see this example (which doesn't compile without type annotations):
```haxe
class Test {
		function foo <T>(x:T->T->T, array:Array<T>) {
    }
    foo ( function (a,b) return a+b, ["foo"]);
	}
}
```

ousado:
Type annotations are essential in certain cases due the relatively limited top-down type inference that Haxe offers, especially when relying on it for the most part. Further, type annotations are extremely useful in combination with the fine-grained control over type conversions that abstracts provide, and they can prevent inference of more general structural types that can lead to slower field access via the respective structural subtyping implementation on currently most static targets.

There's no reason to use type annotations for anyone who doesn't see the benefit. But imposing artificial restrictions like disallowing them on everyone else requires much better arguments than what has been brought forward here.

montibbalt:
We've used short lambdas _extensively_ via a build macro for a few years now and ultimately we had to add support for type annotations by borrowing the type checking syntax. We don't use annotations in the vast majority of cases (I'd guess >98%), but they can be very helpful when necessary. Haxe's type inference can easily get confused and broken on things that have an inheritance hierarchy for example, like the Flash display list, and having the type annotation provides a nice hint to the compiler.  
Here's a few examples of our existing syntax:
Standard short lambda: ``x => x``
Lambda with type annotation: ``(x:Foo) => x``
Curried lambda: ``x => y => x + y``
Multi-argument lambda: ``[x, y] => x + y``
Multi-argument lambda with one argument annotated and curly braces:
```
[(x:Foo), y] => {  
    var z = getSomething();  
    x + y + z;  
}
```

frabbit:
In how many cases did you actually needed to add explicit return type annotations.

montibbalt:
@frabbit for the _return_ type, not enough to bother adding support to our macro. Usually one or two annotations in the arguments is enough to fix most inference issues. I think in those rare cases where we need more, we've either wrapped the return value itself in a type check or broken down and just begrudgingly used the full anonymous function syntax. 
The one mostly-unrelated return type issue I see with relative frequency has to do with a top-level enum that has a constructor with the same name as the enum itself; it can get confused whether you mean the type or the value but this is curiously fixed by just adding the import.

fullofcaffeine:
Is this ever going to be accepted?

ncannasse:
I'm a bit bothered by the non parenthesized cases:
- if we allow multiple parameters such as `a -> b -> c` then it can be read in different ways: `(a) -> ((b) -> c)` or `(a,b) -> c`, so we should not allow this.
- I don't see the value of making optional syntax in the only case of a single parameter given that for 0 or 2+ parenthesizes are mandatory

RealyUniqueName:
I don't see how that can be read in different ways.
`a -> b -> c` is always `a -> (b -> c)` ("old"-style: `function(a) return function (b) return c`), because `->` always means start of a function body.

RealyUniqueName:
As for the second statement, requirement for parenthesis with zero args is obvious, and requirement for parenthesis for multi-args is needed to distinguish new syntax in various lists (function calls, array declarations etc.)

ncannasse:
@RealyUniqueName sorry I was looking at the examples that were posted before your changes by @ousado here https://gist.github.com/ousado/6ab6664144dabe87c66156735920156f

ncannasse:
For the parenthesizes, we agree they are necessary for both 0 and 2+ arguments, I just don't see then a valid reason to make them optional in the specific case of 1 argument

RealyUniqueName:
It's just more convenient for users to omit parentheses when possible.
If that can lead to some syntax related problems in future, then i'm ok with making parentheses mandatory for all cases.

RealyUniqueName:
It's also easier to read:
```haxe
someCall().do(x -> (a  + b) * x).more(x -> x.call());
//vs
someCall().do((x) -> (a  + b) * x).more((x) -> x.call());
```

montibbalt:
If we look at a handful of Haxe's targets with special lambda syntax, it seems like the ones similar to this proposal **do** make the parentheses optional for 1 argument (I would also guess it's for readability). For example: [C#](https://msdn.microsoft.com/en-us/library/bb397687.aspx#Expression Lambdas), [Java](http://docs.oracle.com/javase/tutorial/java/javaOO/lambdaexpressions.html#syntax), and [JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) docs all specifically call it out as optional.  
Other targets do things that are not advisable IMHO. [Python](https://en.wikipedia.org/wiki/Anonymous_function#Python) doesn't require parentheses for the arguments at all, but it also uses a long special keyword. [C++](https://msdn.microsoft.com/en-us/library/dd293608.aspx) looks like it does require them, but it also completely misses the point!  
Personally it seems like a distraction whether the parentheses are optional or not. I do subscribe to the idea that the thing you do the most should have the least amount of syntax, but I don't think making 2 characters optional as a special case (or not) is the thing that should block _this_ particular feature

fponticelli:
I agree with @RealyUniqueName that the optional parenthesis for one argument functions increase readability quite a lot. I hope this will be accepted soon. Can't wait to use it.

frabbit:
I agree with @RealyUniqueName and @fponticelli, the one argument case will be the most used, so it's important that it's short and readable.

ousado:
I also agree - @RealyUniqueName's example demonstrates very nicely how making parenthesis optional in the single argument case helps to avoid visual clutter.

fponticelli:
Now what? Are there more arguments that need to be discussed? How is the decision going to be made?

nadako:
Following the process described in the README, we're at step 4. It means we should either have consensus or voting within haxe core team. I guess we need +1s from @ncannasse @Simn @hughsando @waneck @andyli to know for sure. I'm +1 on that, of course (as well as optional-single-arg-parens bit).

I have a question to @RealyUniqueName about the "drawbacks" section. Is it still relevant? If we decided that `x -> x` is translated to `EFunction({args: [x], expr: EReturn(EIdent(x))})` that shouldn't be any different to current `function(x) return x` to macros, right? If so, please, remove that text. :)

waneck:
I'm in favor of it as well

RealyUniqueName:
@nadako Done!

hughsando:
FWIW, I'm happy to go with it

Simn:
I'm fine with it as well under the condition that we don't come across syntactic ambiguities nobody thought of. Should that happen I reserve the right to change my vote.

RealyUniqueName:
ping @andyli @ncannasse

ncannasse:
I took some time to think about it.

My original opinion on the matter haven't changed much since the last time it was discussed. For me, a nicely designed language is not when there's nothing more to add to it, but when there's nothing more to remove from it, and arrow functions are just a small syntactic sugar that is not necessary for Haxe language. 

Of course, that can be said for other features as well, such as some tricky metadata that we are adding sometimes in the compiler. But these does not need to be understood/learnt by the users to work. My main concern with Array functions is still to introduce some syntax that is confusing by the majority of (new) Haxe users.

However, things have evolved in the recent years, FP has become more and more popular in particular due to the async aspect of JS and Node, which I'm happy with, being myself a long time FP user "before it was cool" (which makes me a FP hipster!). The arrow function has been adopted by more and more languages and is hopefully used by more and more users (which I still think represents a minority of each language users).

In that regard, I'll vote in favor of the current proposal, with the same restrictions that Simon stated. I hope that this does not open a door from various requests to add X or Y new cool syntactic feature recently adopted by cool new language Z, as this is definitely not the way I want Haxe to head in the next years.

One request though, that in order to prevent errors by users mistaking `=>` for `->` we add an error when unifying a function with a `e1 => e2` expression, such as "The operator => is used for maps, did you mean -> for arrow function?"

andyli:
Hey, sorry being late to join this discussion. I've read through the whole proposal and the discussion, and I think the current proposal is promising.

I'm quite interested in the idea @back2dos [made in his comment](https://github.com/HaxeFoundation/haxe-evolution/pull/8#issuecomment-258272236): requiring arrow functions to be pure. It justifies the addition of arrow function pretty well since it will mean that arrow function is not just a syntactic sugar, but actually a proper way to declare pure function, that can be checked by the compiler and potentially enable some optimization.

Similarly, JS's arrow function is not just a syntactic sugar either. The `this` inside a JS arrow function [will always be the instance of the enclosing "class"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#No_binding_of_this), unlike the `this` in classic JS functions can be re-bound to anything using [`func.call(o)` or `func.apply(o)`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/this#call_and_apply).

However, I don't know how much optimization can this guarantee of purity bring. May be @Simn can advice?
If purity is required, may be we have to design a way to specify a pure function type, as I can imagine someone would want to utilize that, for example, to write a function that returns a [memoized function](https://en.wikipedia.org/wiki/Memoization) of an input function.

I know this require some consideration and probably will delay the addition of arrow function, which is what everyone want to use now. But I think this is an interesting idea that deserve some attention. It wouldn't be easy to add/remove this purity requirement later since it would be a breaking change.

RealyUniqueName:
I don't think purity requirement for short lambda is a good thing. What would majority of users do when they see "Your lambda is not pure" error? They will get annoyed :)

In case of JS and `this` lambdas are just fixing language design problem, which Haxe does not have.

fullofcaffeine:
So, what's the next step? :)

andyli:
Look like there is not much interest in the purity requirement after all. I cast my +1 now so we can just move forward. :)

clemos:
To me purity is something entirely different, and it's probably a pity to restrain arrow function usage to pure FP.

back2dos:
I will point out once again, that this proposal gives these two reasons as motivation:

1. Other languages have it.
2. It saves characters.

None of these have ever been reasons to add to the language. I don't see why it should change. This proposal adds *nothing* to the language. The discussion mentions FP as a motivation several times. The proposal doesn't.

As for the discussion, half the posts are pure statements of opinion and don't even try to provide any testable argument. The posts that do propose an objective observation are either superficial or wrong or both. I don't have the time to pick them apart and I bet you have better things to do then to read through me doing it, so let me just pick the last one that tries to make an argument:

> In case of JS and `this` lambdas are just fixing language design problem, which Haxe does not have.

The `this` rebinding in JavaScript is what's required to have OOP. Arrow function expressions cannot be used to declare methods. Now if what's actually being said here is that JavaScript's design is fundamentally flawed (which I wouldn't protest), then I wonder why we should be imitating it.

As for purity:

> I don't think purity requirement for short lambda is a good thing. What would majority of users do when they see "Your lambda is not pure" error? They will get annoyed :)

People get annoyed all the time. Some people are even annoyed that Haxe has a type system to start with. There are more than enough tools out there for them to blow their foot off. There's more value in providing an alternative than in mindlessly imitating the mainstream.

Purity is a very powerful quality, that the analyzer already leverages for optimization. It also guarantees thread safety and composability. And last but not least we can trivially lift this requirement if we wind up with something useless. OTOH introducing it later on will potentially break a lot of code. At the same time no case has actually be made for impure arrow functions. All of the examples given in the proposal are pure. How about some tangible problem to illustrate the point rather than speculating about how some imaginary majority might react?

-----

My reservation though has little to do with purity or not. This proposal has no worthy goal at all. People don't want FP, then that's fine by me. Let's have something else then. But this? What's the manual going to say on the usage of this syntax? "Use this syntax if you wish to save characters or if you need to show your colleagues that Haxe's development is not in stagnation"? What do we tell users who want to use this syntax for method declaration? Will we deprecate the current function syntax if both are equivalent? Etc. etc.

I am all for having a shorter function syntax. But I think it should fit into some bigger picture. If there is one here, it eludes me.

Why care? Why not just add it? Well, I would just point out that we can't use `=>` because adding `OpArrow` seemed such a great idea a few years back (that said `OpArrow` makes it possible to express things in a way that didn't exist before and thus [prevents a certain class of bugs](http://try.haxe.org/#5455B)). This is an opportunity for us to go back and reevaluate the wisdom of the current map literal syntax which not only became an obstacle for this addition, but also does not support empty maps (which is arguably a superficial problem but still smells like hell). No, we just continue. It is short-sighted and too high a price to pay for mere convenience. 

Another case illustrating the cost of such additions is the shorter but otherwise completely redundant `ETernary` which prevents us from having `?` as a unary operator, which could be used for example to allow for things like `nullable?.prop`, `nullable?[index]`, `nullable?(...args)` or `nullable? <op> other` or some other concise syntax to deal with nullness. Just as a reminder: it also prevented us to just use the natural `$e : $t` for `ECheckType` and forced us to go with `($e : $t)` instead, which on top of it all gets parsed as `EParens(ECheckType(e, t))`.

All syntax comes at a cost, so it better be worth it. And I think the cost/benefit ratio can only be sustainable if new syntax also adds new semantics. This proposal doesn't and that's why I will plead one last time that it should either be reframed or rejected. But I guess I'm talking against a wall anyway. So let's just hope this won't become an obstacle when we actually want to add something meaningful to the language. Good luck to us all ;)

RealyUniqueName:
@back2dos Actually the main reason for this proposal is described in the first bullet of "Motivation" part of this proposal.
E.g. for our project short lambdas will dramatically improve readability, thus reducing cost of code review and maintenance.

clemos:
I don't remember these other language features being discussed before.
If they had been discussed in enough details, then maybe people would have foreseen these issues you mentionned, and fixed them.
I think that's what we should do with the proposal at hand to avoid future flaws as much as possible.
This said, I don't think the proposal needs more sophisticated motivations than "it's a shorter way to define anonymous functions, which goes along particularly well with functional and asynchronous programming".

There are already quite a lot of issues raised by this proposal, regardless of the issues it will bring on its own; most of them seem related to how map literals, array comprehension and check-type have been implemented.
I agree these current issues probably need to be fixed before it's even possible to propose some clean and sustainable short lambda design.

back2dos:
@RealyUniqueName "readability" is highly subjective. You might as well say "because it is good". It doesn't mean anything. In fact Nicolas used "readability" for years as an argument *against* short lambdas.

Simn:
The decision has been made, so let's not filibust this issue. We will discuss the exact schedule for this internally and announce it here. For now, I'll go ahead and merge this as an accepted proposal.

markknol:
@back2dos I personally have no insight in what impact syntax changes have but I do trust our skilled compiler team, who all agreed on it. They see no obstacles of adding it in the way it is presented right now, otherwise I would expect them to flag the things you've raised. You are not talking to a wall, I do think if they agree on it there is no need to be concerned.

kevinresol:
@back2dos Then at least the Drawback section shouldn't be empty.

0b1kn00b:
I don't think there's anyone, with the possible exception of Heinz, that
has gotten more out of the Haxe compiler than Juraj.

I have the impression that adding short lambdas was a foregone conclusion,
but for my two cents:

There are good reasons to believe that short lambdas won't make a huge
difference to code quality or productivity in Haxe.

There are several macro versions of short lambdas available, some very
stable, with very little use in published libraries.

This makes the point that for the overwhelming majority of Haxe users, the
cost of using (not writing, using) macros is more than the benefit of short
macros.

That being true, why would anyone make the case that short macros are a
priority? Where a more powerful system that we already have:
metaprogramming, is being avoided *in actual use, *because of other issues
that are being wholesale ignored in this thread.

It doesn't add up, and saying so is not filibustering.

regards,
Laurence

Simn:
You are free to discuss the merits of this, but the decision has been made and will not be overturned this way. That's why I think the energy is best spent elsewhere. Can't make everyone happy.

ousado:
So what's the state of this? 
ping Haxe Core Team @andyli @hughsando @nadako @ncannasse @Simn @waneck

back2dos:
@ousado https://github.com/HaxeFoundation/haxe/issues/5844

nadako:
I think the design is fine and since we already have an implementation proposal, I would say we should move this forward.

Simn:
Would probably help if the implementation proposal was sent as a PR... Preferably with a bunch of tests for "interesting" syntactic cases.

ousado:
Alright. How should those tests look like? Establishing the equivalence of the expected ASTs using macros?

Simn:
That would be nice, but maybe it's too annoying to get right. I'd be content with some pseudo-tests that don't actually assert anything but would fail at typing-time in case something is parsed incorrectly.

ousado:
There you go, let me know if something is missing, I mainly concentrated on the first argument, since that's not reusing `parse_fun_param` in all cases.
https://github.com/HaxeFoundation/haxe/pull/6209

fullofcaffeine:
@ousado __ZOMG__, I can't even believe w're soon going to have short lambdas in Haxe, thanks!

nadako:
It was implemented and available in Haxe 4.0 preview and git version.

shalmu:
Whoa, thanks Dan! Vivat Haxe! 
I bet once 4.0 is in production, we'll see a substantial increase in Haxe popularity.

PR #13, opened 2016-11-11
Create Any_String_For_Abstract_Operator_Overloading

Justinfront:
Suggesting using `` to couch any string for use as an abstract operator. 

Probably this should be submitted in a more formal mannor by a compiler dev or perhaps there is a better approach but felt that the lack of choice for abstract operators ment that some of thier uses would be confusing forcing users to use symbols that were totally not inline with maths or physics standards.  But with `` it maybe possible to create some really interesting algebra systems that would make Haxe very interesting to Universities that currently use languages like c++ and fortran, or comercial stuff like mathematica.  Certainly it would make a lot of sense for Universities to use haxe abstract maths class with HashLink to explore and express ideas quickly and intuatively.

Maybe it's not a full step to the type of expression the Science community might want from Haxe, but perhaps a starting point along with some of the other additions already made.

Abstract as it stands can't have a suitable cross product symbol which seems a bit of a shame and there will be hundreds of similar disappointments for anyone exploring abstract, my suggestion should improve this.  It's more than icing since it allows custom algebra.


proposals/Any_String_For_Abstract_Operator_Overloading
```diff
@@ -0,0 +1,64 @@
...
+I propose the use of ` to couch operators so that any string could be used, so for instance users could create an `add` function.
+
+Usage:
+```haxe
+abstract MyAbstract(String) {
+  public inline function new(s:String) {
+    this = s;
+  }
+
+  @:op(A `add` B)
```
RealyUniqueName: Only binary operators?

Justinfront:
I have added lines explaining this should be implemented for unary and comparisom operators giving an example use I did not flesh out the abstract construction as it should be obvious instead of a currently exceptable symbol a user can use `what ever she needs` .

nadako:
Please use proper file name for the proposal:
> Fork the repo, copy the 0000-template.md to proposals/0000-short-name.md, where short-name is a descriptive filename for the proposal document. Don't assign the number yet.

Justinfront:
Nadako, had another go but may not have got it quite right.

markknol:
What is the advantage of <code>fred `faster than` tom</code> over <code>fred.fasterThan(tom)</code> ?

frabbit:
How would you deal with Operator precedence?

Justinfront:
Currently with the abstract class your very limited on what you can use for operators, this improvement would allow you to create more intuative algebra.
var crossProduct = a\`×\`b;
var dotProduct = a\`·\`b;
obviously it would be nice if there was also a way to take it even further..
var absValue = \`|\`a\`|\`;

Operator precedence is not actually related to this change because I don't believe it would be in any different from how abstracts operators currently work, I don't believe there is currently a way to define precedence? 
Maybe my suggestion is highlighting a current limitation but that is additional to the proposal ie: something that could be implemented independantly as any implementation would not be different from what would be required to give precedence to the current abstract operators.  I presume the order of declaration is the precedence but have not tested?

Justinfront:
You could use it with complex numbers.
```haxe
var angle = `sin` myComplexNo;
```
with pi
```haxe 
var circle = 2 `π `;
```
I think it allows you to be more expressive with Maths from within Haxe.

frabbit:
well operator precedence is very well related to this issue, because all operators that you can overload now have a well defined precedence. But if you introduce new/custom ones, they need a precedence too.

Justinfront:
Well it seems logical that precedence is related to the order in your class with the highest being the first one you add.

RealyUniqueName:
No it's not )
In Haxe order of fields declaration has no meaning. Let's not change that.
In case this feature will be accepted, it's better to specify prescendence via some special meta.
E.g.
```haxe
@:customOpPrescendence(`√`, `π`, `|`)
abstract MyType...
```

Justinfront:
I don't see that 'custom' is necessary. 
Further 'Prescedence' is quite fancy and easy to spell wrong I think far simpler to use.
```
@:orderOfOperation(`√`, `π`, `|`)
abstract MyType...
```
since they seem to be equal terms, and while pescedence sounds smarter it also less general used word and probably more familiar only with people with math technical background, although it might be french origin - so familiarity may depend on users perspective, but to me 'order of operation' is plain and simple and would be my preference.
https://en.wikipedia.org/wiki/Order_of_operations

kevinresol:
> Well it seems logical that precedence is related to the order in your class with the highest being the first one you add.

Not only within your own list of operators but also with the "standard" operators

For example: ``5 + v1 `dot` v2``,  should it be ``(5 + v1) `dot` v2`` or ``5 + (v1 `dot` v2)``?

ncannasse:
I dont't think that's a good idea. It introduces some very alien syntax in Haxe with not much gain in terms readability or keystrokes vs using a proper method name

Justinfront:
Could we look to add more symbols to the current list and consider how orderOfOperation is defined it seems to be fixed and that seems pretty limited approach.
It would be really nice to be able to write 2π for instance, I am sure there are more symbols that should be possible for use as an operator.

ncannasse:
I think most of the developers (myself included) don't even know how to write a pi with their keyboard. Adding more operators which don't have a clear definition (and pi is not an operator) leads to abusing the operators to create some confusing DSLs

Simn:
This has been rejected, see Nicolas' reasoning.

Justinfront:
More limited request...

1) Can I request orderOfOperation be made customizable it's very crazy that for abstracts it's fixed as it further limits maths sets that can be nicely represented. 

2) Any chance of adding  ' x ' to allow easy cross product use which is common or would this be too tricky to implement?

PR #15, opened 2016-12-26
Add generator expressions from existing array comprehension syntax

binki:
[rendered version](https://github.com/binki/haxe-evolution/blob/generator-expressions/proposals/0000-generator-expressions.md)

Originally pitched at https://github.com/HaxeFoundation/haxe/issues/5909 .

I think that because Array Comprehension syntax exists it would be
natural to have Generator Expressions with the same syntax. However,
if Haxe implements a yield statement, work on that should probably
supercede consideration of this particular proposal.

### DIFF
--- a/proposals/0000-generator-expressions.md
+++ b/proposals/0000-generator-expressions.md
@@ -1,0 +1,266 @@
+# Generator Expressions
+
+* Proposal: [HXP-NNNN](NNNN-generator-expressions.md)
+* Author: [Nathan Phillip Brink](https://github.com/binki)
+
+## Introduction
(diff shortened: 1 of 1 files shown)


Simn:
This proposals focuses on the syntax of a new feature while neglecting the actual implementation it would require. Lazy evaluation is not something you can just switch on. It requires to think about what code should be generated and in our case this means thinking about all the target generators.

Generators can be implemented based on something like async/await. We are currently considering what exactly we want to support in this regard for Haxe 4. I am thus closing this proposal.

PR #18, opened 2017-02-20
Expression macro method forwarding

nadako:
Provide a way to have expression macro functions inside a class without typing that whole class in macro context.

[Rendered version](https://github.com/nadako/haxe-evolution/blob/macro_forward/proposals/0000-macro-forward.md)

### DIFF
--- a/proposals/0000-macro-forward.md
+++ b/proposals/0000-macro-forward.md
@@ -1,11 +1,11 @@
-# Macro method forwarding
+# Expression macro method forwarding
 
 * Proposal: [HXP-NNNN](NNNN-filename.md)
 * Author: [Dan Korostelev](https://github.com/nadako)
 
 ## Introduction
 
-Provide a way to have macro functions inside a class without typing that whole class
+Provide a way to have expression macro functions inside a class without typing that whole class
 in macro context.
 
 ## Motivation
--- a/proposals/0000-macro-forward.md
+++ b/proposals/0000-macro-forward.md
@@ -1,4 +1,4 @@
-# Forwarding macro calls
+# Macro method forwarding
 
 * Proposal: [HXP-NNNN](NNNN-filename.md)
 * Author: [Dan Korostelev](https://github.com/nadako)
--- a/proposals/0000-macro-forward.md
+++ b/proposals/0000-macro-forward.md
@@ -61,7 +61,7 @@ for that metadata and extract the path to a real method to be evaluated. So if t
 real method is in another module, the current one doesn't need to be typed in macro
 context at all.
 
-In additional to per-method macro we could have class-level `@:forwardMacro(MyMacroClass)`
+In addition to per-method macro we could have class-level `@:forwardMacro(MyMacroClass)`
 metadata that will forward all bodyless macro methods from this class to the specified `MyMacroClass`.
 
 ## Impact on existing code
--- a/proposals/0000-macro-forward.md
+++ b/proposals/0000-macro-forward.md
@@ -5,7 +5,7 @@
 
 ## Introduction
 
-Provide a way have macro functions inside a class without typing that whole class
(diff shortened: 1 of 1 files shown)


proposals/0000-macro-forward.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Detailed design
+
+The proposed solution is quite simple: allow macro methods without body if they have
+a special metadata pointing to the static method in another type. For example, the function
+from the example above could be defined like this instead:
+
+```haxe
+@:forwardMacro(MyMacroTools.getVersionExpr)
+static macro function getVersion();
```
RealyUniqueName: Why is this better than calling `MyMacroTools.getVersionExpr()` directly instead of `getVersion()`?

nadako: Because, as stated in the beginning, we often want macro methods to be members of our class, so we could call it like `Main.getVersion()`.

jcward: Gotta say, I agree with @RealyUniqueName -- I don't see the benefit of this forward. It doesn't actually achieve keeping the code in your main class, it just introduces another indirection syntax. Meh.

nadako: It's not about keeping code in the class, it's about providing a way to avoid typing the class with macro + non-macro methods in a macro context.

jcward: > we often want macro methods to be members of our class

> It's not about keeping code in the class

:confused:

nadako: I can understand the confusion and I agree that there might be better ways to deal with the original problem, but this specific proposal is not about addressing beginner's confusion, it's about having a way to have macro methods without typing the class in a macro context, which is currently _not possible at all_.

As @Simn noted elsewhere it _should_ be possible to have macro code coexisting with non-macro code in a single class but that means reworking the typer to provide lazy processing, which I think is a huge change (that I doubt will be done in near future), while the proposed here solution provides a simple way to address problem.

frabbit: @jcward as @nadako  stated, this about a clean separation between macro classes and regular classes. The problem arises when you mix macro and regular methods in a class, in that case that class is typed in both contexts macro and regular which is often/mostly not what you want and you have to add `#if macro` and `#if !macro` all over the place. It does actually what you would expect normally when mixing macro and not macro methods. It's also hard to explain when you never hit a problem like that, but if you write a lot of macro code you can fairly easy generate a problem like that which is very hard to understand especially when you have platform specific types in the same class.

Simn:
I don't really get it. If you implement your macro code in another .hx file anyway, why not simply import the macro function to the calling module instead of doing this forwarding?

```haxe
import MyMacroTools.getVersion;
js.Browser.console.log("Version " + getVersion());
```

nadako:
Well, the example might not display the actual issue where it's not possible - member macro fields. They have to be on the class and thus they'll force typing of the module of that class in macro context.

Simn:
You can use `using` instead of import and add a first argument to your macro.

nadako:
That looks like a workaround for the actual issue, because in my case, those macro methods are by design not an extension, but a proper API of the class. Also I would love to not have to add additional import/usings. In fact, I tried that approach combined with `import.hx`, but run into some build order issue I can't remember right now. I'd say if we have member macro methods, they should somehow work without the pain of fencing the rest of the code in the module (which is also error-prone as mentioned in the proposal).

Simn:
> That looks like a workaround for the actual issue

So does your proposal. :P

nadako:
Well that's true :), but I imagine implementing that is much easier and wouldn't affect other stuff. If you have a better solution then I don't mind closing this proposal as outdated.

Simn:
https://github.com/HaxeFoundation/haxe/issues/7453

PR #24, opened 2017-07-18
Module-level functions

nadako:
Allow defining functions (and maybe vars) directly in .hx files instead of making a class with static methods.

[Rendered version](https://github.com/nadako/haxe-evolution/blob/module_level_funcs/proposals/0000-module-level-funcs.md)

### DIFF
--- a/proposals/0000-module-level-funcs.md
+++ b/proposals/0000-module-level-funcs.md
@@ -41,7 +41,7 @@ Here's some code to break down the wall of text a bit. A slightly complicated he
 
 Hello.hx
 ```haxe
-inline var USAGE = "Usage: hello <name> [times]";
+inline var USAGE = "Usage: hello <name> <times>";
 
 typedef Config = {
     name:String,
--- a/proposals/0000-module-level-funcs.md
+++ b/proposals/0000-module-level-funcs.md
@@ -62,6 +62,8 @@ function main() {
 }
 ```
 
+As you can see, mixing module-level functions and vars with other type declarations (`typedef Config` here) works just fine.
+
 ### Reflection
 
 Since the module-level functions/vars end up being static class fields, the usual reflection should automatically work (e.g. `Reflect.field(Type.resolveClass("MyModule"), "myMethod")`). Actually I'm not sure if this is specified to work currently in Haxe, but if it does, generators should respect that and don't over-optimize `KModuleStatics` generation when reflection features are enabled.
--- a/proposals/0000-module-level-funcs.md
+++ b/proposals/0000-module-level-funcs.md
@@ -68,9 +68,16 @@ As you can see, mixing module-level functions and vars with other type declarati
 
 Since the module-level functions/vars end up being static class fields, the usual reflection should automatically work (e.g. `Reflect.field(Type.resolveClass("MyModule"), "myMethod")`). Actually I'm not sure if this is specified to work currently in Haxe, but if it does, generators should respect that and don't over-optimize `KModuleStatics` generation when reflection features are enabled.
 
-## Macros, static extensions and so on
+### Macros, static extensions and so on
 
 Since, again, module-level functions/vars are actually static fields, usual rules for `using` and `@:build(MyModule.myMethod)` calls apply, nothing new here.
+
+### Final note
+
+Because this document proposes implementing module-level functions in form of static class fields,
+one may think that it would be inconsistent to have them public and imported implicitly with the module, remember however that this proposal is not about providing syntax sugar for static fields, but about defining functions and var on module level, and the fact they become static fields is an _implementation detail_.
+
+Accessibility and resolution should follow those of other module-level declarations, that is: a module-level declaration (e.g. class/typedef/etc) is public unless specified as private, all module-level declarations are imported when the module is imported. Following these rules it makes sense to have module-levels functions to be public and be imported by default.
 
 ## Impact on existing code
 
--- a/proposals/0000-module-level-funcs.md
+++ b/proposals/0000-module-level-funcs.md
@@ -88,7 +88,12 @@ I don't see any viable alternatives that would allow defining plain functions. H
 
 ## Opening possibilities
 
-I think we could also use module-level functions to define extern functions (e.g. `extern function SDL_Init(flags:UInt32):Int`), but that's gonna require some more thought, because it would mean that the implicitly created class must be made extern automatically.
+I think we could also use module-level functions to define extern functions, e.g.
+```haxe
+extern function SDL_Init(flags:UInt32):Int;
+````
+
+But that's gonna require some more thought, because it would mean that the implicitly created class must be made extern automatically.
 
 ## Unresolved questions
 
--- a/proposals/0000-module-level-funcs.md
+++ b/proposals/0000-module-level-funcs.md
@@ -41,25 +41,24 @@ Here's some code to break down the wall of text a bit. A slightly complicated he
 
 Hello.hx
 ```haxe
-inline var USAGE = "<name> [times]";
+inline var USAGE = "Usage: hello <name> [times]";
 
-function sayHello(name) trace('Hi, $name!');
+typedef Config = {
+    name:String,
+    times:Int,
+}
+
+function sayHello(config:Config) {
+    for (i in 0...config.times)
+        trace('Hi, ${config.name}!');
+}
 
 function main() {
-    var name, times;
-    switch Sys.args() {
-        case [n]:
-            name = n;
-            times = 1;
-        case [n,t]:
-            name = n;
-            times = Std.parseInt(t);
-        default:
-            trace(USAGE);
-            return;
-    }
-    for (_ in 0...times)
-        sayHello(name);
+    var args = Sys.args();
+    if (args.length != 2)
+        trace(USAGE);
+    else
+        sayHello({name: args[0], times: Std.parseInt(args[1])});
 }
 ```
 
--- a/proposals/0000-module-level-funcs.md
+++ b/proposals/0000-module-level-funcs.md
@@ -39,10 +39,10 @@ We have two options to deal with this:
 
  1) Implicitly create primary module class for module level functions and forbid explicit primary module class in this case. This would be the safest and most minimal change.
  2) Allow both primary module class and module statics class and apply additional identifier resolution rules:
-   * `import MyModule` imports both types and module-level functions from `MyModule` to local namespace.
-   * `import MyModule.method` if there's a module-level `method` function - import it, otherwise look for `method` in the primary class statics as it's done now.
-   * `import MyModule.*` imports both module-level functions and primary class statics, module-levels taking precedence.
-   * `MyModule.method()` calls the module-level function, if found, otherwise calls static method of the primary class.
+    * `import MyModule` imports both types and module-level functions from `MyModule` to local namespace.
+    * `import MyModule.method` if there's a module-level `method` function - import it, otherwise look for `method` in the primary class statics as it's done now.
+    * `import MyModule.*` imports both module-level functions and primary class statics, module-levels taking precedence.
+    * `MyModule.method()` calls the module-level function, if found, otherwise calls static method of the primary class.
 
 I propose going with the first option at least for now, because that would greatly simplify implementation and won't introduce new resolution semantics. Moreover, I think it'll be what people actually want when using module-level function declarations.
 
--- a/proposals/0000-module-level-funcs.md
+++ b/proposals/0000-module-level-funcs.md
@@ -11,7 +11,7 @@ Support defining functions directly in the module (.hx file) instead  of creatin
 
 Classes are the heart of object-oriented programming found in languages like Java. In this paradigm, we model our program as interaction between class instances, so it makes sense to have classes as containers for everything. Non-factory static methods are relatively rare, so there's no practical need for functions defined outside of a class.
 
-Nowadays, however, other programming paradigms, such as functional programming is becoming more and more popular, reducing the need for classic OOP classes and instead focusing on functions that process passive data structures.
+Nowadays, however, other programming paradigms, such as functional programming are becoming more and more popular, reducing the need for classic OOP classes and instead focusing on functions that process passive data structures.
 
 Haxe provides a lot of features supporting functional oriented paradigms (most importantly first-class functions), however it lacks a clean way to actually define functions without creating a wrapping class. This is annoying and gives a feeling of bloatedness to new people coming from non-OOP background. This is particularly unfortunate, because most of our target languages support plain functions, so having to wrap everything in a class can be a con when deciding whether to use Haxe or a target language directly.
 
--- a/proposals/0000-module-level-funcs.md
+++ b/proposals/0000-module-level-funcs.md
@@ -1,11 +1,11 @@
-# Module-level functions
+# Module-level functions and variables
 
 * Proposal: [HXP-NNNN](NNNN-module-level-funcs.md)
 * Author: [Dan Korostelev](https://github.com/nadako)
 
 ## Introduction
 
-Support defining functions directly in the module (.hx file) instead of creating a class with static methods.
+Support defining `function`s and `var`s directly in the module (.hx file) instead of creating a class with static fields.
 
 ## Motivation
 
@@ -21,34 +21,23 @@ Supporting module-level functions should be pretty-straightforward. To minimize 
 
  * add `TDFunction(name:String, fun:Function)` case to the `TypeDefKind` enum.
  * allow parsing functions at the module level and parse them into that `TDFunction`.
- * on module loading, when processing syntax declarations into module types, treat all module-level functions as static methods of an
-   implicitly created class. For this class we introduce a new `ClassKind` variant: `KModuleStatics` or
-   something. This is very similar to how `KAbstractImpl`-classes are implicitly created for abstract types.
+ * on module loading, when processing syntax declarations into module types, treat all module-level functions as `public static` methods of an implicitly created class. For this class we introduce a new `ClassKind` variant: `KModuleStatics` or something. This is very similar to how `KAbstractImpl`-classes are implicitly created for abstract types.
  * thereafter, when resolving an identifier (see more below), actually generate a static field access (`TTypeExpr(ModuleStatics).static(name)`).
- * when generating output, if target supports declaring plain functions (JavaScript, Lua, C++, etc.), a generator can decide to lose the `KModuleStatics` class and generate functions directly. If target requires a wrapping class (Java, C#) - generate like a normal class (plus, some optimizations an be applied, e.g. C# could mark class as `static`, and don't generate reflection helpers).
+ * when generating output, if target supports declaring plain functions (JavaScript, Lua, C++, etc.), a generator can decide to lose the wrapping `KModuleStatics` class and generate functions directly. If target requires a wrapping class (Java, C#) - generate like a normal class (plus, some optimizations an be applied, e.g. C# could mark class as `static`, and don't generate reflection helpers).
 
-> While this proposal describes module-level functions, I think it would be logical and consistent (although less useful) to also allow module-level variables, treating them similarly static vars.
+While the default access modifier for module-level functions is `public`, private module-level functions are supported by explicitly specifying the `private` keyword.
 
-The `-main` argument should also handle module-level functions for program entry points.
+While this proposal describes module-level functions, vars are implemented the same with a `TDVar` variant. Properties can be supported too.
 
 ### Identifier resolution
 
-The idea of module-level identifiers doesn't play particularly well with our current static field resolution mechanism, mainly because we already have the concept of "primary module class" (a class with the same name as the module), so we have to think about the sane resolution rules.
+The idea of module-level identifiers doesn't play particularly well with our current static field resolution mechanism, because we already have the concept of "primary module class" (a class with the same name as the module), so the most logical and least intrusive way to implement this would be to simply collect module-level functions in a implicitly created primary class and forbid explicit primary class declaration when there are module-level functions or vars.
 
-We have two options to deal with this:
-
- 1) Implicitly create primary module class for module-level functions and forbid explicit primary module class in this case. This would be the safest and most minimal change.
- 2) Allow both primary module class and module statics class and apply additional identifier resolution rules:
-    * `import MyModule` imports both types and module-level functions from `MyModule` to local namespace.
-    * `import MyModule.method` if there's a module-level `method` function - import it, otherwise look for `method` in the primary class statics as it's done now.
-    * `import MyModule.*` imports both module-level functions and primary class statics, module-levels taking precedence.
-    * `MyModule.method()` calls the module-level function, if found, otherwise calls static method of the primary class.
-
-I propose going for the first option at least for now, because that would greatly simplify implementation and won't introduce new resolution semantics. Moreover, I think it'll be what people actually want when using module-level function declarations.
+Importing a module with `KModuleStatics` primary class should also imply `import ModuleName.*`, which means that all module-level functions/vars are imported by importing the module, which I believe would be the expected behaviour.
 
 ### Code example
 
-Here's some code to break down the wall of test a bit (also including module-level variables for the sake of example):
+Here's some code to break down the wall of text a bit. A slightly complicated hello-world command line script would look something like this:
 
 Hello.hx
 ```haxe
@@ -74,6 +63,14 @@ function main() {
 }
 ```
 
+### Reflection
+
+Since the module-level functions/vars end up being static class fields, the usual reflection should automatically work (e.g. `Reflect.field(Type.resolveClass("MyModule"), "myMethod")`). Actually I'm not sure if this is specified to work currently in Haxe, but if it does, generators should respect that and don't over-optimize `KModuleStatics` generation when reflection features are enabled.
+
+## Macros, static extensions and so on
+
+Since, again, module-level functions/vars are actually static fields, usual rules for `using` and `@:build(MyModule.myMethod)` calls apply, nothing new here.
+
 ## Impact on existing code
 
 With regard to existing code, this change can only potentially affect macro code because of newly introduced enum constructors in the macro API, and I believe that a very small portion of macro code will be affected by this, because it only matters for exhaustive pattern matches on `TypeDefKind` and `ClassKind` which are quite rare.
@@ -97,5 +94,4 @@ But that's gonna require some more thought, because it would mean that the impli
 
 ## Unresolved questions
 
- * importing/identifier resolution - two variants proposed.
- * module-level vars - generally considered bad style, but would be consistent to have and can actually be useful for small scripts
+None so far.
--- a/proposals/0000-module-level-funcs.md
+++ b/proposals/0000-module-level-funcs.md
@@ -1,0 +1,81 @@
+# Module-level functions and variables
+
+* Proposal: [HXP-NNNN](NNNN-module-level-funcs.md)
+* Author: [Dan Korostelev](https://github.com/nadako)
+
+## Introduction
+
+Support defining functions directly in the module (.hx file) instead
+of creating a class with static methods.
+
+## Motivation
+
+Classes are the heart of object-oriented programming found in languages like Java. In this paradigm, we
+model our program as interaction between class instances, so it makes sense to have classes as containers
+for everything. Non-factory static methods are relatively rare, so there's no need for functions defined outside of a class.
+
+Nowadays, however, other programming paradigms, such as functional programming is becoming more and more
+popular, reducing the need for classic OOP classes and instead focusing on functions that process passive
+data structures.
+
+Haxe provides a lot of features supporting functional oriented paradigms (most importantly first-class functions),
+however it lacks a clean way to actually define functions without creating a wrapping class. This is annoying and
+gives a feeling of bloatedness to new people coming from non-OOP background. This is particularly unfortunate,
+because most of our target languages support plain functions, so having to wrap everything in a
+class can be a con when deciding whether to use Haxe or a target language directly.
+
+## Detailed design
+
+Supporting module-level functions should be pretty-straightforward. To minimize changes in compiler and its data
+structures, as well as the macro API, I propose the following:
+
+ * add `TDFunction(name:String, fun:Function)` case to the `TypeDefKind` enum;
+ * allow parsing functions at the module level and parse them into that `TDFunction`;
+ * when processind syntax into typed AST, treat all module-level functions as static methods of an
+   implicitly created class. For this class we introduce a new `ClassKind` variant: `KModuleStatics` or
+   something. This is very similar to how `KAbstractImpl`-classes are implicitly created for abstract types.
+ * thereafter, when resolving an identifier (see more below), actually generate a static field access (`TTypeExpr(ModuleStatics).static(name)`)
+ * when generating output, if target supports declaring plain functions (JavaScript, Lua, C, etc.), a generator
+   can decide to lose the `KModuleStatics` class and generate functions directly. If target requires a wrapping
+   class (Java, C#), generate like a normal class (plus, some optimizations an be applied, e.g. C# could mark
+   class as `static`, and don't generate reflection helpers).
+
+While this proposal describes module-level functions, I think it would be logical and consistent (although less useful) to also allow module-level variables, treating them similarly static vars.
+
+The `-main` argument should also handle module-level functions for program entry points.
+
+### Identifier resolution
+
+The idea of module-level identifiers doesn't play particularly well with our current static field resolution mechanism, mainly because we already have the concept of "primary module class", so we have to think about the sane resolution rules. Here's what I propose:
+
+ * `import MyModule` imports both types and module-level functions from `MyModule` to local namespace.
+ * `import MyModule.method` if there's a module-level `method` function - import it, otherwise look for `method` in the primary class statics as it's done now.
+ * `import MyModule.*` imports both module-level functions and primary class statics, module-levels taking precedence.
+
+fully-qualified paths are resolved accordingly:
+
+ * `MyModule.method()` calls the module-level function, if found, otherwise calls static method of the primary class.
+
+> TODO: Maybe we should just forbid having explicit primary type if there are module-level functions and generate
+it implicitly. This should cover most use-cases and greatly simplify resolution rules (basically, no changes to it will be required at all).
+
+## Impact on existing code
+
+> TODO: only macro data structures, not changed but added, might only break exhaustive pattern matches.
+
+## Drawbacks
+
+> TODO: Describe the drawbacks of the proposed design worth consideration. This doesn't include breaking changes, since that's described in the previous section.
+
+## Alternatives
+
+> TODO: What alternatives have you considered to address the same problem, why the proposed solution is better?
+
+## Opening possibilities
+
+> TODO: Does this change make other future changes possible or easier? Leave this section out if the proposed change
+is completely self-contained.
+
+## Unresolved questions
+
+> TODO: Which parts of the design in question is still to be determined?
--- a/proposals/0000-module-level-funcs.md
+++ b/proposals/0000-module-level-funcs.md
@@ -5,29 +5,19 @@
 
 ## Introduction
 
-Support defining functions directly in the module (.hx file) instead
-of creating a class with static methods.
+Support defining functions directly in the module (.hx file) instead  of creating a class with static methods.
 
 ## Motivation
 
-Classes are the heart of object-oriented programming found in languages like Java. In this paradigm, we
-model our program as interaction between class instances, so it makes sense to have classes as containers
-for everything. Non-factory static methods are relatively rare, so there's no practical need for functions defined outside of a class.
+Classes are the heart of object-oriented programming found in languages like Java. In this paradigm, we model our program as interaction between class instances, so it makes sense to have classes as containers for everything. Non-factory static methods are relatively rare, so there's no practical need for functions defined outside of a class.
 
-Nowadays, however, other programming paradigms, such as functional programming is becoming more and more
-popular, reducing the need for classic OOP classes and instead focusing on functions that process passive
-data structures.
+Nowadays, however, other programming paradigms, such as functional programming is becoming more and more popular, reducing the need for classic OOP classes and instead focusing on functions that process passive data structures.
 
-Haxe provides a lot of features supporting functional oriented paradigms (most importantly first-class functions),
-however it lacks a clean way to actually define functions without creating a wrapping class. This is annoying and
-gives a feeling of bloatedness to new people coming from non-OOP background. This is particularly unfortunate,
-because most of our target languages support plain functions, so having to wrap everything in a
-class can be a con when deciding whether to use Haxe or a target language directly.
+Haxe provides a lot of features supporting functional oriented paradigms (most importantly first-class functions), however it lacks a clean way to actually define functions without creating a wrapping class. This is annoying and gives a feeling of bloatedness to new people coming from non-OOP background. This is particularly unfortunate, because most of our target languages support plain functions, so having to wrap everything in a class can be a con when deciding whether to use Haxe or a target language directly.
 
 ## Detailed design
 
-Supporting module-level functions should be pretty-straightforward. To minimize changes in compiler and its data
-structures, as well as the macro API, I propose the following:
+Supporting module-level functions should be pretty-straightforward. To minimize changes in compiler and its data structures, as well as the macro API, I propose the following:
 
  * add `TDFunction(name:String, fun:Function)` case to the `TypeDefKind` enum;
  * allow parsing functions at the module level and parse them into that `TDFunction`;
@@ -35,10 +25,7 @@ structures, as well as the macro API, I propose the following:
    implicitly created class. For this class we introduce a new `ClassKind` variant: `KModuleStatics` or
    something. This is very similar to how `KAbstractImpl`-classes are implicitly created for abstract types.
  * thereafter, when resolving an identifier (see more below), actually generate a static field access (`TTypeExpr(ModuleStatics).static(name)`)
- * when generating output, if target supports declaring plain functions (JavaScript, Lua, C, etc.), a generator
-   can decide to lose the `KModuleStatics` class and generate functions directly. If target requires a wrapping
-   class (Java, C#), generate like a normal class (plus, some optimizations an be applied, e.g. C# could mark
-   class as `static`, and don't generate reflection helpers).
+ * when generating output, if target supports declaring plain functions (JavaScript, Lua, C, etc.), a generator can decide to lose the `KModuleStatics` class and generate functions directly. If target requires a wrapping class (Java, C#), generate like a normal class (plus, some optimizations an be applied, e.g. C# could mark class as `static`, and don't generate reflection helpers).
 
 While this proposal describes module-level functions, I think it would be logical and consistent (although less useful) to also allow module-level variables, treating them similarly static vars.
 
@@ -46,22 +33,23 @@ The `-main` argument should also handle module-level functions for program entry
 
 ### Identifier resolution
 
-The idea of module-level identifiers doesn't play particularly well with our current static field resolution mechanism, mainly because we already have the concept of "primary module class", so we have to think about the sane resolution rules. Here's what I propose:
+The idea of module-level identifiers doesn't play particularly well with our current static field resolution mechanism, mainly because we already have the concept of "primary module class" (a class with the same name as the module), so we have to think about the sane resolution rules.
 
- * `import MyModule` imports both types and module-level functions from `MyModule` to local namespace.
- * `import MyModule.method` if there's a module-level `method` function - import it, otherwise look for `method` in the primary class statics as it's done now.
- * `import MyModule.*` imports both module-level functions and primary class statics, module-levels taking precedence.
+We have two options to deal with this:
 
-fully-qualified paths are resolved accordingly:
+ 1) Implicitly create primary module class for module level functions and forbid explicit primary module class in this case. This would be the safest and most minimal change.
+ 2) Allow both primary module class and module statics class and apply additional identifier resolution rules:
+   * `import MyModule` imports both types and module-level functions from `MyModule` to local namespace.
+   * `import MyModule.method` if there's a module-level `method` function - import it, otherwise look for `method` in the primary class statics as it's done now.
+   * `import MyModule.*` imports both module-level functions and primary class statics, module-levels taking precedence.
+   * `MyModule.method()` calls the module-level function, if found, otherwise calls static method of the primary class.
 
- * `MyModule.method()` calls the module-level function, if found, otherwise calls static method of the primary class.
-
-> TODO: Maybe we should just forbid having explicit primary type if there are module-level functions and generate
-it implicitly. This should cover most use-cases and greatly simplify resolution rules (basically, no changes to it will be required at all).
+I propose going with the first option at least for now, because that would greatly simplify implementation and won't introduce new resolution semantics. Moreover, I think it'll be what people actually want when using module-level function declarations.
 
 ## Impact on existing code
 
-> TODO: only macro data structures, not changed but added, might only break exhaustive pattern matches.
+With regard to existing code, this change can only potentially affect macro code because of newly introduced enum constructors in the macro API, and I believe that a very small portion of macro
+code will be affected by this, because it only matters for exhaustive pattern matches on `TypeDefKind` and `ClassKind` which are quite rare.
 
 ## Drawbacks
 
--- a/proposals/0000-module-level-funcs.md
+++ b/proposals/0000-module-level-funcs.md
@@ -46,6 +46,34 @@ We have two options to deal with this:
 
 I propose going for the first option at least for now, because that would greatly simplify implementation and won't introduce new resolution semantics. Moreover, I think it'll be what people actually want when using module-level function declarations.
 
+### Code example
+
+Here's some code to break down the wall of test a bit (also including module-level variables for the sake of example):
+
+Hello.hx
+```haxe
+inline var USAGE = "<name> [times]";
+
+function sayHello(name) trace('Hi, $name!');
+
+function main() {
+    var name, times;
+    switch Sys.args() {
+        case [n]:
+            name = n;
+            times = 1;
+        case [n,t]:
+            name = n;
+            times = Std.parseInt(t);
+        default:
+            trace(USAGE);
+            return;
+    }
+    for (_ in 0...times)
+        sayHello(name);
+}
+```
+
 ## Impact on existing code
 
 With regard to existing code, this change can only potentially affect macro code because of newly introduced enum constructors in the macro API, and I believe that a very small portion of macro code will be affected by this, because it only matters for exhaustive pattern matches on `TypeDefKind` and `ClassKind` which are quite rare.
@@ -57,6 +85,10 @@ I don't immediately see any drawbacks in the proposed feature. On the contrary, 
 ## Alternatives
 
 I don't see any viable alternatives that would allow defining plain functions. Having a Haxe superset that is compiled to Haxe with a macro or in any other way isn't something anyone would seriously consider in practice.
+
+## Opening possibilities
+
+I think we could also use module-level functions to define extern functions (e.g. `extern function SDL_Init(flags:UInt32):Int`), but that's gonna require some more thought, because it would mean that the implicitly created class must be made extern automatically.
 
 ## Unresolved questions
 
--- a/proposals/0000-module-level-funcs.md
+++ b/proposals/0000-module-level-funcs.md
@@ -5,13 +5,13 @@
 
 ## Introduction
 
-Support defining functions directly in the module (.hx file) instead  of creating a class with static methods.
+Support defining functions directly in the module (.hx file) instead of creating a class with static methods.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


Justinfront:
So the proposal is to allow stuff like this?

```haxe
package;
// import Main.hello;
class Main {
  static public function main():Void {
    hello();
  }
}
function hello(){
  trace('hello function');
}
```

Could main even be just a function if user wanted is that part of your proposal?

```haxe
package;
function main():Void { 
   hello();
}
function hello(){
  trace('hello function');
}
```

So essentially in more general case this:

```haxe
package myStuff;
class MyFunctions{
   static function hello(){
      trace('my function hello');
   }
}
```

could be written:

```haxe
package myStuff;
// package myStuff.MyFunctions;
function hello(){
      trace('my function hello');
}
```

Interesting proposal if I have understood but seems kind of like icing for the user especially if it worked like below, your just kind of removing the 'class' word.

```haxe
package myStuff.MyFunctions;
function hello(){
      trace('my function hello');
}
```

but then you could not also put a class in the file, and could you have two groups of functions in a file and import them separately surely the extra of a class makes little odds?  Is it not acceptable if no constructor to allow users to drop using static accept that I guess via reflection you could still create a class so you would need different holder name.

```haxe
package myStuff;
functions MyFunctions{
   function hello(){
      trace('my function hello');
   }
}
```

Which seems a lot of change just so that users can avoid word 'static' and looks pretty the same as now.

I think the proposal could do with samples of how code might look atleast in your prefered solution, it would make it clearer for people who struggle to read technical definitions but are fine seeing code patterns.

So would constants or variables be in these packages ?

```haxe
package myStuff;
const helloText = 'hello';
function hello(){
  trace( helloText );
}
function hello2(){
 trace( helloText + ' ' + helloText );
}
```

nadako:
> Could main even be just a function if user wanted is that part of your proposal?

Yes, I mentioned that. In fact, it'll automatically work if the implicitly created class for functions will be the primary module class.

>  your just kind of removing the 'class' word

That's the intention - when you don't do OOP, this class wrapping becomes just a visual clutter. And it's not just the `class` keyword. It's also class name duplicating the module name, braces, `static` keyword for each function, extra indentation level etc.

> but then you could not also put a class in the file

That's not true. This proposal allow for placing module-level functions/vars along with other types just fine, that's the idea.

> would constants or variables be in these packages ?

it's mentioned in the proposal as well, that's one of the open questions, but i think it wouldn't hurt to support that as well.

> the proposal could do with samples of how code might look

thanks, I added a sample. however you seem to have read the proposal carelessly, because most of what you're asking about was mentioned there... maybe it's because of my english though. :)

benmerckx:
Allow marking those as private as well? I think `TDFunction(name:String, fun:Function)` wouldn't allow for that as `Function` doesn't have the needed `constraints`?
```haxe
private function rand() return Math.random();
var MY_RANDOM_NUMBER = rand();
```

nadako:
> Allow marking those as private as well? I think TDFunction(name:String, fun:Function) wouldn't allow for that 

Yes, private functions should totally be supported, though it just occured to me that haxe.macro.Expr structures doesn't have a notion for other `private` types either, so I guess that's something that should be added to the `TypeDefinition` structure instead of `TypeDefKind` constructors.

frabbit:
Maybe the easiest way to implement this is to put all top level functions as statics automatically in the module named class and disallow the definition of a class named as the module itself. This way it would only be nice sugar for this common case.

nadako:
> Maybe the easiest way to implement this is to put all top level functions as statics automatically in the module named class and disallow the definition of a class named as the module itself. This way it would only be nice sugar for this common case.

Yes that's what I proposed. I also propose a more complex option, but I think we should settle with this.

Justinfront:
The example helps thank you. Your english is fine, code examples really are much clearer than words for some of us, even if they don't cover all the detail.

I don't really understand what frabbit said maybe a specific example would help.

Looking at the example you added, I am still not clear on how you would prefer/imagine package/import will work in practice, the example glosses over thier use, could you add an example of your prefered package and import use.  For instance if I had a file   MyMath, and you have a group of functions called trig, and a group called matricies in the MyMath file, and you wanted to use one in one class and one in another, what would that actually look like, or is that not supported?

I understand the proposal is for functional approaches to be more clear, but it seems it is just sugar. I don't really see it adding much beyond creating 2 ways to do one thing, which will make learning more complex overall - even if initially it's simpler, your adding a lot of new - and it all needs explaining, even something like lots of slightly different ways of doing 'if' add to the amount people need to learn before they can be productive. 

The proposal could make for cleaner haxe code and make it clearer that Haxe has functional leanings, but it's hard for me to see it as more than a conceptial icing change, I don't see the bigger picture of how this is more than just reducing keystrokes maybe that's enough but not historically!

Aurel300:
Interesting. It is basically syntactic sugar, but so was the arrow notation for functions and people were pretty happy with that. Some more issues should probably be clarified:

 - How do you treat static extensions?

Typically e.g. `using Lambda;` would look through all `public static function`s in the class `Lambda` in the module `Lambda`. With your proposal, does this include the classless functions? As I understand it, they are implicitly static.

```
// Lombdo.hx
function double(x:Float) return x * 2;
```

```
// Main.hx
using Lombdo;
static function main() trace(Math.PI.double());
```

Should this compile?

 - Macro context?

Can the module-level functions be declared `macro`? How do you use them as build macros?

 - Reflection?

Is there any way to access the module-level functions of a module with reflection? Currently the reflection and type APIs can only handle classes and instances, I don't think they have a notion of modules. Is a function declared at module level a function instance? What is its lifetime?

markknol:
This would make explaining Haxe and documenting it much easier, any most many examples don't need classes and could run this way. Making small tools would also easier since a main function and maybe some other function can already be a full Haxe application. Its not that big deal to wrap class around it, but if you can leave it out, why not.

My only concern is that I think you directly want module-level variables too, otherwise that will confuse.

Also try.haxe could remove it's "simple" mode, which now wraps everything in a class with static main function. @clemos

nadako:
@Aurel300 since the idea is to place module-level functions in an implicitly created class (so they end up being its static methods), they are going to work just fine as macros and static extensions. and reflection is a good point in favor of making that class the primary module class (so you `resolveClass` the module name and access its statics as usual).

clemos:
@markknol: on try.haxe, it's indeed just a matter of shortening code samples to a minimum, not at all about advocating for a more "open" way to define functions or variables.
I must say I'm quite used to the current way (incl. with `import MyStaticHelpers.*`) and find it convenient as is.
Actually, I think that introducing this could pollute the toplevel namespace, potentially importing a lot of (unwanted) toplevel identifiers.
Also, this would add yet another complexity level to identifier resolution.
Finally, I'm not sure I understand how this proposal relates to functionnal programming, other than making it easier to avoid the-Java sounding `class` keyword :trollface:

Aurel300:
@clemos I can see the appeal functional programming. A lot of times, I would like to do something like `arr.map(flatten)` or `zip(arr1, arr2)` or `arr.map(id)`, etc. `Lambda` in std. does something similar, when you use it as a static extension. However, because it is a static extension, it will only ever be available as a method of a object, i.e. you have to access it with the dot operator. With FP-like functions it is often needed to be able to access them anywhere, without having an instance – because, after all, you might want to pass a function to a higher-order function. So compare:

```
arr.map(a -> a.flatten())
```

And:

```
arr.map(flatten)
```

Other functions, like `zip`, don't really make sense when accessed as a method of one of its two arguments.

So yes, importing a module with a bunch of functions like this would "pollute" your namespace, although naturally any variables or functions you define locally override the imported ones. However, there are situations when you are writing very FP-focused code and it would be beneficial (for legibility and nice syntax) to actually have them in the namespace.

fullofcaffeine:
Hey guys, can we proceed with this one? I think we have enough 👍 for it to be merged.

markknol:
The core team should decide here.

Pauan:
This is one of the few things preventing me from switching from TypeScript to Haxe, so this gets a **big** :+1: from me.

> Haxe provides a lot of features supporting functional oriented paradigms (most importantly first-class functions)

In addition to first-class functions, Haxe also has ADTs (e.g. enums) and pattern matching, which is *very* unusual for an OOP language (but definitely a good thing!)

Haxe definitely has the potential to be an excellent multi-paradigm language, supporting both OOP and functional style seamlessly. It just needs a little nudge in the right direction (e.g. this proposal).

> While this proposal describes module-level functions, I think it would be logical and consistent (although less useful) to also allow module-level variables, treating them similarly static vars.

If functions are "elevated" into static methods of an implicit class, couldn't top-level `var` be elevated into static members of the implicit class?

I think top-level `var` will absolutely be needed: to define module state (e.g. a counter):

```haxe
var counter: Int = 0;

public function increment(): Int {
    return ++counter;
}
```

Or to define static data structures which are shared between functions:

```haxe
var foo = {
  // various properties defined here
};

public function bar() {
  // do stuff with foo
}
```

nadako:
Yes, I mentioned toplevel `var` in the original proposal and I think it should be done as well. I also now think that it should simply become an implicitly created main module class with static fields. I'll update the proposal tomorrow to be more clear about it and maybe provide an sample implementation.

fullofcaffeine:
As per your proposal, we will have a way to execute code at the module? We will be able to call functions at the module-level?

nadako:
@fullofcaffeine if you're talking about arbitrary module-level expressions, then no, it's only about defining module-level functions and vars

fullofcaffeine:
Fair enough, I guess module-level expressions would just overcomplicate things, since we would have to assume code would be executed upon importing modules, which could lead to unwanted side-effects. I think the proposal is good as it is.

back2dos:
> Haxe provides a lot of features supporting functional oriented paradigms (most importantly first-class functions), however it lacks a clean way to actually define functions without creating a wrapping class. **This is annoying and gives a feeling of bloatedness to new people coming from non-OOP background.**  [emphasis added]

Everyone with a decent understanding of functional programming will consider nesting functions in a class as a minor nuisance at best. Even Scala requires global functions to be defined [on some object](https://stackoverflow.com/questions/18548814/can-i-define-and-use-functions-outside-classes-and-objects-in-scala). People drawn to functional programming will choose Scala over Haxe for a pretty long list of reasons, none of which is about having to type a keyword more or less.

The benefits of this proposal seem negligible to me, as opposed to the complexity added. If you want to do FP practitioners a favor, that's much appreciated, but this just ain't it ;)

ibilon:
@back2dos maybe not for people coming from functional programming yeah
but from imperative languages or even any language which has top level function/expression it makes porting code a lot easier. For instance from javascript.

> as a minor nuisance at best

It certainly is only minor, but for people trying out haxe as an alternative among others it can be the detail that drives them away. Especially since they are more likely to try some small code, and top level function makes hello world level snipet a lot succinct.

back2dos:
> It certainly is only minor, but for people trying out haxe as an alternative among others it can be the detail that drives them away. Especially since they are more likely to try some small code, and top level function makes hello world level snipet a lot succinct.

People who will evaluate a language based on hello world snippets and such minute details should not be taken seriously, and certainly not be catered to. I seriously doubt we can expect any meaningful contribution to our ecosystem from them. 

On the other hand long time contributors like Massive Interactive and Franco are leaving Haxe behind. Those are the people we should not be losing:

- for Massive it was probably the friction in JS interop, which I think could have been most effectively tackled by consuming .d.ts files from DefinitelyTyped ... I find it hard to imagine that it'd be much harder than the jar/dll parsing in hxcs and hxjava. Pulling that off will also certainly be far more interesting to JS people considering Haxe than having top level functions.
- for Franco it was the lack of FP features. There's a whole list of them, most of which would probably be quite the challenge to implement. However, immutability would be relatively easy, but it's something that especially Nicolas refuses to see any value in.

There are many measures we could take to make Haxe more appealing to frontend web devs and FP followers (two groups with a rapidly growing intersection). This is not one of them. That's no reason not to add this, but I wanted to make it as clear as I can what this proposal will not be achieving, regardless of hopes projected onto it. If the remaining benefits are considered worth the efforts, by all means go ahead ;)

Pauan:
@back2dos Scala is not representative of functional programming: every other functional language does *not* have nested classes, instead they use top-level functions. Scala uses nested classes because it needs to interop with Java.

In addition, **every** non-OOP language allows for top-level vars and functions: Python, Ruby, JavaScript, TypeScript, Perl, PHP, every Lisp, C, C++, F#, Haskell, OCaml, PureScript, Elm, etc.

It's not just a "functional language" thing, it's an "every non-OOP language" thing. It's far more common for a language to have top-level vars and functions than it is to *not* have them.

I can guarantee you that many people from non-OOP languages (including JavaScript and TypeScript) will be very turned off by requiring functions to have an extra indentation level and `static`.

These people are not useless, they are simply used to having a less verbose language, so why would they choose to switch to a more verbose language?

Adoption rates are controlled by two things: the benefits of switching and the costs of switching. There are many benefits to switching to Haxe, but there is also a lot of costs (including this issue). Removing the costs tends to increase adoption rates more so than increasing the benefits.

That's why TypeScript is so successful: it's so easy to switch from JavaScript to TypeScript. People don't care that TypeScript is a bad language. They care about how easy it is to switch.

That's not illogical or stupid behavior: switching languages is a very big deal. Learning a new language requires a **huge** time and energy investment, and rewriting a program in a new language is *also* a **huge** time and energy investment. Languages are a long-term investment. People are only going to switch if they think it's worth it.

Making it easy to switch to another language reduces the time and energy investment, which makes switching much more attractive.

And people aren't going to spend months evaluating whether a language is good enough to switch: they're going to make that decision quickly, usually after taking a look at some examples and the tutorial. There's so many languages out there, and so little time to choose, they can't afford to spend a lot of time carefully analyzing a language.

Catering to beginners increases adoption rates dramatically, because even people who have been programming for decades in other languages are still a *beginner to Haxe*. Beginner doesn't just mean "new to programming" or "bad programmer", it also means "good programmer who is used to other languages".

There are so many good languages to choose from, why would people choose a language which has a bad first-time experience? Why would people choose a language which forces them to write their code more verbosely for no benefit?

When you are writing a large program, you will have thousands of functions, and if every single function needs an extra indentation level and an extra `static`, that's very annoying.

Remember: because languages are a long-term investment, if somebody chooses to use Haxe, they will be stuck with the extra verbosity *forever*. It's not just a few characters. It's a few characters multiplied by every function, multiplied by every program, multiplied for years. It's not just a "minor" thing that only affects newbies.

nadako:
@back2dos off-topic, but since I researched d.ts loading, I have to comment on this for future readers:

Consuming .d.ts files is actually much harder than parsing strictly structured low-level .net/java bytecode: TS has high-level stuff like declaration merging, type queries, type unions/intersections, literal types and so on. We would be reimplementing a big part of TS compiler, which is, unlike .NET/Java bytecode is constantly evolving. Even if we had resources to support that, automatic externs generated from d.ts would be way worse than ones that come from jar/dll. The proper way to go here is to write a d.ts-to-extern converter that uses TS compiler through its API do do the typing heavy-lifting. Then, if we make it flexible enough to be usable without changing externs by hand (much), we could look into integrating it with Haxe compiler more tightly.

nadako:
@back2dos on-topic, while I was writing my previous comment @Pauan already said most of what I had in mind :) I'll just add my 2 cents here:

 * I was trying not to use the actual "functional programming" term much, not to trigger FP-related discussions :)
 * One can write a large application in a non-oop way in Haxe using just functions and anonymous structures, and these `class`/ `public static`/indent stuff becomes useless clutter that makes code looks worse and more annoying to write. this was one of my point in the "Motivation" section
 * I think Haxe can be a great language to learn programming from scratch. Everyone is trying to teach using JS nowadays, but since it's too quirky and dynamic, it doesn't help at all when one is using it wrongly (tried on my wife^^). If we don't have to introduce the concept of `class` right away, Haxe would be as concise (if not more) as JS, plus compile-time checking that gives hints that are actually useful to a newbie, minus all the weirdness - it would be a nice first language.
 * Knowing some compiler internals, I'm pretty sure it's easy and non-invasive to implement, so even if it's not a HUGE deal, it's still something to consider, given the reasoning above in this thread.

Justinfront:
Having thought about Dan's proposal some more I am in favour, it makes lots of code cleaner and simpler.  Typescript is only relevant to Haxe in terms of Angular's popularity and if Haxe manages to promote a better framework then that could disappear tomorrow.  Chasing Typescript won't allow Haxe to become popular.  Haxe's main problem is that still few people of heard of it let alone considered it.  This change makes Haxe more approachable and more feasible for it to be taught as a first language in schools.

back2dos:
First of all, let me address that this whole notion of using Haxe for teaching is outrageously preposterous

back2dos:
First of all, let me address this whole idea that this will allow using Haxe for teaching, which is outrageously preposterous:

1. This proposal will not, in the slightest, allow to write class-less scripts in Haxe. For that to be the case, one would actually have to be able to put arbitrary statements on the top level, which is a big deal. The point being that while `trace("hello world")` is a valid [try.haxe](https://try.haxe.org/#CAbdD) snippet, it's not in and on itself a valid Haxe program. Making it so is arguably a mildly interesting idea, but it's a whole other deal than what's proposed here.
2. Putting together a good learning environment is a tremendous undertaking. Having one that is better than JavaScript might be possible. But to think that trivial changes like these will bring us even anywhere near Smalltalk, Self, Lisp (with Light Table while you're at it), Apparatus or the gazillion of other well engineered *interactive* environments is outright stupid and an insult to decades of creative research into the field. A REPL would maybe be a start to get anywhere near that.
3. The mandatory `class` keyword does not stop a significant portion of universities to use Java as their first programming language. It's not a hurdle for beginners. It's not a hurdle for experience programmers. If anything, it's a turn-off for petty-minded, superficial nitpicks. Among them are the kind of trolls who run around the JS community, bullying everyone who does not comply to their narrow vision of things. We'll do just fine without them.

Secondly: Haxe has always erred on the side of explicitness. If that is to change, then great (I guess). But there still should be a *coherent* way to drive language design forward (like if we can omit the class, can we also omit the package statement, which is fully redundant?), rather than uncoordinatedly cramming in party tricks here and there. The premise is fine, but the proposed solution falls short and may well prove an obstacle in adopting a design that addresses the problem.

Lastly: the Spoon language has all kinds of neat syntactic toys. Project got abandoned because nobody really cares at the end of the day. [The same fate is dawning upon CoffeeScript](https://stateofjs.com/2016/flavors/). I am all for streamlining our syntax to our usecases. But I do fail to see the use case here. Some indentation is not that big a deal. As pointed out, Scala has the same restriction and while I have never claimed it to be representative of anything, it still ranks as [the most proliferated FP language on GitHub](https://octoverse.github.com/). In light of that, the claim that having to nest your code one extra level (as for example in AS3, where everything is *in* a package statement for no good reason) is a critical impediment to adoption falls flat.

Pauan:
@back2dos 

> This proposal will not, in the slightest, allow to write class-less scripts in Haxe.

You can write class-less scripts if you define a top-level `main` function.

>  But to think that trivial changes like these will bring us even anywhere near Smalltalk, Self, Lisp (with Light Table while you're at it), Apparatus or the gazillion of other well engineered interactive environments is outright stupid and an insult to decades of creative research into the field.

**Nobody** claimed that this change *by itself* will lead to massive adoption. **Nobody** claimed that this change *by itself* will suddenly make Haxe an excellent teaching language. You are the only one claiming that.

The only claim is that this change moves things in the right direction. It nudges things closer to where they should be.

If your criteria for judging proposals is "it has to fix **all** of the problems **all at once** (including unrelated problems)", then nothing will ever get done.

Improvements have to be made one step at a time. This proposal is one of those steps. Nobody here believes that this is the only step, and nobody here believes that it is a big step. But it is a step.

> A REPL would maybe be a start to get anywhere near that.

That's a great idea! And it's also completely unrelated to this proposal. You are free to make your own proposals if you wish.

> Secondly: Haxe has always erred on the side of explicitness.

All of the top-level vars and functions are fully explicit. The only implicit part is the creation of the class.

Haxe already implicitly lets you leave off the class name if the class name is the same as the module name, so this proposal is consistent with Haxe's module system.

> like if we can omit the class, can we also omit the package statement, which is fully redundant?

No you cannot omit the `package` keyword, because that actually conveys necessary information. How would you propose that we remove it?

The `static` keyword on the other hand does not convey necessary information, it's just extra noise.

> It's not a hurdle for beginners. It's not a hurdle for experience programmers.

It is a hurdle for some beginners, it is a hurdle for some experienced programmers.

If that was not true, then why do you think the Try Haxe site allows for top-level vars, functions, and expressions?

The fact that Try Haxe is inconsistent with the rest of the language causes problems, and you claim that there is *no benefit* to top-level vars and functions, so why do you think the creators of Try Haxe chose to allow top-level vars and functions *despite* the problems?

> rather than uncoordinatedly cramming in party tricks here and there

I'm not sure why you're claiming this is "uncoordinated": this is a valid proposal which follows all of the rules and protocol of haxe-evolution.

It doesn't suddenly open the flood-gates to millions of other syntax proposals, because every proposal is judged independently on its own merits.

And the core Haxe team will of course consider this proposal carefully when they vote on it. There's absolutely *nothing* "uncoordinated" about it.

> proposed solution falls short and may well prove an obstacle in adopting a design that addresses the problem

I do not see how this proposal creates an obstacle for anything, could you clarify?

And what alternate proposal would you have for people who like having top-level vars and functions (which is the vast majority of programmers)?

> The same fate is dawning upon CoffeeScript.

You pointed to failed languages to try to prove your point, but you ignore all of the *successful* languages.

CoffeeScript failed because ES6 became popular, and one of the reasons ES6 became popular is *because* of having nice syntax sugar.

*Every* language has syntax sugar. The differences are in how much syntax sugar they have, and what kind of syntax sugar they have.

Claiming that allowing top-level vars and functions is "too much syntax sugar and will cause your language to fail" is absurd.

The vast overwhelming majority of *successful* languages have top-level vars and functions.

The vast overwhelming majority of programmers are used to top-level vars and functions.

This isn't a "JavaScript vs Haxe" thing, it's not a "functional programming vs Haxe" thing, it's a basic language thing.

> Lastly: the Spoon language has all kinds of neat syntactic toys. Project got abandoned because nobody really cares at the end of the day.

The truth is that language popularity is really complicated, and it's often unclear why a language failed or succeeded.

You claimed that Spoon failed because nobody cares about syntax sugar, but you have presented *no evidence* for that claim. The burden of proof is *on you* to back up your claims.

> As pointed out, Scala has the same restriction

Scala caters to *Java programmers* who are already used to using classes for everything. Scala is popular because *Java is already popular*.

I very seriously doubt that Scala is appealing to non-Java programmers.

> In light of that, the claim that having to nest your code one extra level (as for example in AS3, where everything is in a package statement for no good reason) is a critical impediment to adoption falls flat.

Nobody claimed it is a *critical impediment*, just that it's an impediment.

Some people will consider it a major impediment, other people (such as yourself) will consider it not an impediment at all. People have different preferences.

If your goal is to appeal to Java programmers, then the mandatory class isn't an impediment at all.

If your goal is to appeal to web developers and functional programmers, then yes it is a rather significant impediment, because those types of programmers are used to having top-level vars and functions.

If you don't believe me that it is an impediment, go talk to some web developers and non-Scala functional programmers, show them some Haxe code and tell them that the class is mandatory, and that *all* functions and vars must have the `static` keyword. See what their reaction is.

But of course when you see their reaction, you'll just dismiss them as "petty-minded superficial nitpick trolls who should be dismissed because they will never contribute to Haxe". Despite the fact that they *do* contribute to the languages that they are currently using.

Imagine that Haxe 5 was released, and one of the syntax changes is that now all methods must have a new `method` keyword:

```haxe
class Foo {
  public method function foo(): Int {
    return 5;
  }
}
```

This new `method` keyword is mandatory on *all* methods, and it serves no useful purpose.

Would you accept that? Would you think that is okay?

And if some existing Haxe users said "screw that, we're never going to upgrade to Haxe 5, we're staying with Haxe 4 forever", would you call them "petty-minded superficial nitpick trolls who will never contribute anything to Haxe"?

From the perspective of non-Haxe users, *that's exactly the situation they're in*: they are used to having top-level functions and vars, and now you are telling them that they *must* add in this useless `static` keyword to *every* var and function.

> it's a turn-off for petty-minded, superficial nitpicks. Among them are the kind of trolls who run around the JS community, bullying everyone who does not comply to their narrow vision of things. We'll do just fine without them.

Please stop claiming that the vast majority of programmers should "be dismissed because they will not contribute to Haxe" and are "petty-minded superficial nitpick trolls".

It's completely fine if you don't like the proposal.

It's completely fine if you think the proposal is useless.

It's completely fine if you disagree with the proposal.

But please stop insulting everybody who disagrees with you.

Not everybody who has different preferences than you is a moron or a troll.

A large part of language design (especially syntax) is subjective and people will disagree. That's okay, not everybody can agree on everything, but that also means that *other people aren't wrong just because they disagree with you*.

Pauan:
I would like to make something very clear:

Mandatory classes are an impediment for some people. Therefore this proposal will increase adoption rates.

But we don't know how much: will it increase it by 0.1%? 1%? 5%? 10%? 50%?

We have not done any interviews, any polls, any studies, therefore the only correct answer is "we don't know"

In addition, this proposal does not mention adoption rates.

Therefore, this proposal should be judged based on its own merits:

* It significantly shortens the code of Haxe programmers who prefer an imperative or functional style

* It significantly shortens the code of static extensions

* It significantly shortens the code of macros

* It significantly shortens the code of Haxe shell scripts

* It removes some inconsistencies between Try Haxe and the real language (this is a problem that does trip up people new to Haxe)

nadako:
@Pauan again said everything I had to reply in a more comprehensive way than I possibly could. :)

I can only add that I indeed don't see much of a point in `package` statements, since packages always strictly follow directory structure and they didn't ever do anything for me other than being nuisance when moving modules around. But that's a completely different topic for another evolution proposal.

Anyway, I made some edits to the proposal, answering some questions and going with a simple primary-module-class approach.

Pauan:
@nadako It seems a little inconsistent to me that module-level functions and vars are automatically `public`

I wouldn't mind having them be private by default, but I don't have strong opinions either way.

----

Auto-importing the functions and vars is also inconsistent.

How do you propose that Haxe programmers should fix name collisions if two different modules create top-level functions/vars with the same names?

nadako:
I also added more to-the-point code sample that illustrates that module-level functions works with other type declarations.

nadako:
@Pauan 

> It seems a little inconsistent to me that module-level functions and vars are automatically public

You might be right about this, but I also don't have a strong opinion here.

> Auto-importing the functions and vars is also inconsistent.

I disagree, because when you import a module, you import all its declaration into your module namespace, so if we have module-level function declarations it's only consistent to import them as well.

> How do you propose that Haxe programmers should fix name collisions if two different modules create top-level functions/vars with the same names?

Like they do that now, by reordering imports or writing more specific import (`import MyModule.concreteMethod`, `import MyModule.concreteMethod as differentMethod`, etc.)

nadako:
> You might be right about this, but I also don't have a strong opinion here.

On the second thought - I have: normally, types declared in module are public unless explicitly private, so this should apply to module-level functions and vars too.

Pauan:
@nadako 

> I disagree, because when you import a module, you import all its declaration into your module namespace, so if we have module-level function declarations it's only consistent to import them as well.
>
> On the second thought - I have: normally, types declared in module are public unless explicitly private, so this should apply to module-level functions and vars too.

You're right, but at the same time `static` fields are automatically `private`, and top-level functions and vars are indeed `static` fields, so it's a bit odd.

After thinking about it more, I suppose from the programmer's perspective top-level functions/vars are more similar to top-level enums/classes than `static` fields, so having auto-`public` and auto-import is fine.

> Like they do that now, by reordering imports or writing more specific import (`import MyModule.concreteMethod`, `import MyModule.concreteMethod as differentMethod`, etc.)

Okay, great, I don't have any more concerns with the proposal.

nadako:
> from the programmer's perspective top-level functions/vars are more similar to top-level enums/classes than static fields

Yes, the fact they become static fields is an implementation detail.

Pauan:
@nadako You might want to put an explanation for why they are auto-`public` and auto-import, since the reader's first instinct is to think that it's inconsistent with `static` fields.

----

Also, you mentioned that the code generator can optimize it by not creating the implicit class. You also mentioned that optimization won't work with reflection.

However, it is possible for the code generator to hoist all static fields (vars and functions) to the top-level.

So for example, this Haxe code:

```haxe
class Foo {
  public static function foo(): Int {
    return bar();
  }

  public static function bar(): Int {
    return 5;
  }
}
```

Would get compiled into this JavaScript code:

```js
function foo() {
  return bar();
}

function bar() {
  return 5;
}

function Foo() {}
Foo.foo = foo;
Foo.bar = bar;
```

Because the fields are hoisted, it can dispatch to them directly (it calls `bar()`, not `Foo.bar()`), but reflection still works because the fields exist on `Foo`

And if reflection isn't used, the dead code eliminator can then remove the `foo` and `bar` fields (and also the `Foo` class if it doesn't have any fields left).

So in the common case it would only output the `foo` and `bar` functions, and would not output the `Foo` class.

This optimization should be possible for *all* `static` fields for *every* class, so it would automatically apply to top-level vars and functions as well. No special rules needed!

nadako:
Good catch @Pauan, thanks!

Pauan:
I thought about it some more, and my optimization idea is trickier than I thought: it could still cause things to break if `Reflect.deleteField`, `Reflect.setField`, or `Reflect.setProperty` are used.

However, it's possible to fix the breakage by using ES5 getters/setters.

But maybe we don't need to fix it, because [the `Reflect` API makes no guarantees about correctness](https://haxe.org/manual/std-reflection.html): it even specifically says that the `Reflect` API might not work because of dead code elimination.

In any case, this is starting to get off-topic, so this optimization idea should be moved to another issue.

My point was just that we can optimize *all* `static` fields of *all* classes, so we don't need special optimization rules for top-level vars/functions.

nadako:
> You might want to put an explanation for why they are auto-public and auto-import, since the reader's first instinct is to think that it's inconsistent with static fields.

Done.

fullofcaffeine:
Sooooo, should we merge this? 🙄

nadako:
> Sooooo, should we merge this now? 🙄

Well unfortunately none of the core developers expressed their opinions yet. Also after looking into implementing it a bit, I'm considering lifting the "primary module class" restriction, because I now think supporting both primary module type and module-level functions/vars wouldn't be THAT hard.

fullofcaffeine:
Can we ping them? I feel that evolution docs, even after ready, take a lot of time to be approved/merged because of that. A good habit might be to always ping all the core team members here when it's ready for review.

nadako:
Well, sure: @ncannasse @Simn @hughsando @jdonaldson @andyli @waneck

waneck:
I really like this

jdonaldson:
I'd be willing to do the work to support this for lua.  But, the way I see it I shouldn't have to do anything at all... so maybe I'm not getting the whole picture.

Is it really necessary to define new AST nodes for this?  Why not just infer the class name from the file name, generate a wrapper class in the AST for it, and add special metadata to it to indicate that it behaves as a module level class.   Even for targets that have no notion of class, it will still be important to namespace the code so that methods and variables do not collide.   "Pure" module level functions won't really generate faster code either... most targets sensitive to this have flattened namespaces.

Also, can we reliably differentiate between class-level and field-level metadata for this?  Some metadata applies to both.

As a side note, it would be nice to handle non-oop class constructs with this (e.g. abstract/extern)

jdonaldson:
Also, kudos to @nadako for all the documentation work on the proposal.  Very succinct and clear.

nadako:
> Is it really necessary to define new AST nodes for this? Why not just infer the class name from the file name, generate a wrapper class in the AST for it, and add special metadata to it to indicate that it behaves as a module level class. Even for targets that have no notion of class, it will still be important to namespace the code so that methods and variables do not collide.

What I'm proposing is to have a separate `TypeDefKind` for module-levels in the untyped AST (stuff represented by `haxe.macro.Expr` structures), because untyped AST basically represents syntax and there's no class in the syntax. While when it's typed into the typed AST (stuff represented by `haxe.macro.Type` structures), I indeed propose to create a wrapper class. Then a generator can choose to omit the wrapping class or not.

> Also, can we reliably differentiate between class-level and field-level metadata for this? Some metadata applies to both.

Well since there's no class for module-levels in the syntax, one can't add class-level metadata for them. I think that's a non-issue, because it's rarely needed and it's still possible to create a normal class with static methods and metadata.

> As a side note, it would be nice to handle non-oop class constructs with this (e.g. abstract/extern)

Not sure I understand what you mean...

hughsando:
I love the idea of a top-level main function/other functions.  The details I will leave up to you.

andyli:
I think the title and focus should be "module-level properties (i.e. with get/setter) and methods" instead of "module-level variables and functions".

The proposal mentioned `Properties can be supported too.`, but really, I think it *should* be supported. 

In Haxe, currently,

 * On expression level, we have only var and function (no prop), and other expressions. The evaluation order is the same as the order of how they are written. There is no access modifier (public/private).

 * On class level, we have var, prop, and method. The evaluation (initialization) order is determined by their dependencies. There are access modifiers.

 * On module level, we have types (class, interface, enum etc.). The initialization order is determined by their dependencies. There are access modifiers.

As you may have noticed, the module level is more similar to the class level than the expression level. That means promoting things from the class level (instead of the expression level) to the module level can be more appropriate. And I can see that this proposal is already talking about the class level var and function instead of the expressions. I see no obvious reason to leave properties behind.

RealyUniqueName:
Add module-level `main` function or better add module-level expressions and you got shell scripting covered )

nadako:
> Add module-level main function or better add module-level expressions and you got shell scripting covered )

Module-level `main` function will be automatically possible, but expressions is too much, Haxe is not a scripting language :)

hughsando:
Haxe is not a scripting language is a bit recursive because I think it
would actually be quite good in this niche.
Doing something like "var _ = () -> { script goes here; }()" is possible,
so disallowing expressions just seems a missed opportunity.

andyli:
Adding module-level expressions requires a few more decisions because it can be confusing. For example, a `var a = "a";` in the top-level can be a module-level variable **or** a module-level expression... and they are different.

If all `var` are module-level variables, one can write something that is invalid in plain expression, e.g.
```haxe
trace(a);
var a = "a";
```
which is similar to:
```haxe
class A {
    static var _ = { trace(a); null; }
    static var a = "a";
}
```

RealyUniqueName:
Maybe there should be a EBlock for module-level exprassions?
```haxe
package my.pack;

{
   //module-level expressions allowed here
}
```

andyli:
> a EBlock for module-level exprassions?

Yeah, this is one solution, but then there is more decisions, e.g. Can there be multiple blocks? What is the evaluation order and output of the following?

```haxe
package test;

{
    trace(a);
    a = 456;
}
var a = 123;
class Test {
    static function main() {
        trace(a);
    }
}
```

andyli:
To clarify, I do like module-level expressions, but if we want to support it, this proposal probably needs a rewrite to add all the necessary details.

nadako:
Yeah, module-level expressions is a whole different topic raising quite a bit of questions, let's not derail this proposal :)

hughsando:
Yes, I guess I'm not really interested in module-level declarations.
I'm more into a "simplified main file", which basically wraps everything in
"class Main{public static function main() {  ...... }}".  Since there would
only be one of these, and no class defs and no external referencing, this
is pretty orthogonal to module issues. And would solve my scripting use
case.

nadako:
> which basically wraps everything in
"class Main{public static function main() {  ...... }}"

Slightly off-topic, but I actually thought about implementing something like `--expr MyFile` that does exactly that, which would be great for learning programming, IMO. Nowadays tho I think we should look into makign some REPL(ish) thing, but I'm not sure if that would be easy to implement in our current architecture.

fullofcaffeine:
>Yes, I guess I'm not really interested in module-level declarations.
I'm more into a "simplified main file", which basically wraps everything in
"class Main{public static function main() {  ...... }}".  Since there would
only be one of these, and no class defs and no external referencing, this
is pretty orthogonal to module issues. And would solve my scripting use
case.

I'd love to see that as well, and I think that saying Haxe is not a scripting language is a false premise. 
 It can be and it is sometimes used for that. I prefer using Ruby or Python for quick jobs (what I call scripting) because there's less friction, namely, I can just start writing the sequence of commands I need without adding any more structure to support it (like a main class or function for example). 

However, I do think it should be handled in a separate request and ideally after this one is implemented, which is already a step in the right direction in my opinion.

jdonaldson:
>> Also, can we reliably differentiate between class-level and field-level metadata for this? Some metadata applies to both.

> Well since there's no class for module-levels in the syntax, one can't add class-level metadata for them. I think that's a non-issue, because it's rarely needed and it's still possible to create a normal class with static methods and metadata.

I think I understand better how the compiler will process these.  What you say makes sense.  Would it be worth a workaround to introduce module-level metadata nodes that get applied to the class?

>>As a side note, it would be nice to handle non-oop class constructs with this (e.g. abstract/extern)
>Not sure I understand what you mean...

There's some class-related modifiers that are _not_ related to OOP.  The ```abstract``` and ```extern``` keywords come to mind.  I'm just wondering if we should support those here.

Pauan:
@jdonaldson 

> Would it be worth a workaround to introduce module-level metadata nodes that get applied to the class?

I think that's a good idea, but I think it should be in a separate proposal.

Keeping this proposal lightweight improves the chances that it gets accepted and implemented.

> The `abstract` and `extern` keywords come to mind. I'm just wondering if we should support those here.

With this proposal you can continue to have top-level `abstract` and `extern` just like normal.

If you mean the ability to define `abstract` and `extern` methods/variables/properties at the top-level, I think that should be in a separate proposal.

ncannasse:
Looks ok to me, I would however allow to have both toplevel function/var and one or several classes, maybe reserving `MyModuleScope` for the pseudo class path (for MyModule.hx).  -- we might want to look for MyModuleScope.main for -main as a consequence.

So import MyModule would import all types + imply import MyModule.MyModuleScope.* (if exists and is KModuleStatics).

Having them public by default is normal since it's the same as class/typedef/enum/etc. (they are toplevel declarations). You could have them marked as private as well, which mean that MyModuleScope needs @:allow(MyModule) so that MyModule other classes can access these.

Please remove all mention of reflection for now, we don't want this to be part of the spec :)

extern looks useful, if they are all extern we can make MyModuleScope extern as well, and leave the platform code generator to deal with how to handle things "natively".

nadako:
Glad to see a general approval of this!

I'm going to work on an implementation for a bit, to better understand the level of complexity of supporting both primary module class and module-level functions/vars and to do some refactoring in relevant compiler parts for the greater good :)

After that I'm gonna update this proposal with more detailed implementation details.

> maybe reserving MyModuleScope for the pseudo class path (for MyModule.hx)

I don't like reserving names, I think the better way is to have a private class with mangled name generated similar to how it's done for KAbstractImpl classes.

> You could have them marked as private as well, which mean that MyModuleScope needs @:allow(MyModule) so that MyModule other classes can access these.

Good catch about @:allow!

> Please remove all mention of reflection for now, we don't want this to be part of the spec :)

I'm quite surprised that you don't want to have reflection specified, but that's perfectly fine with me :)

> extern looks useful

I'd say let's leave that for a future proposal.

jcward:
**FYI, off-topic:** there is a Haxe REPL project called `ihx` (seemingly like `irb`, the Ruby REPL): https://github.com/ianxm/ihx

I'm not familiar with the dev, nor have I peeked at how it works, but I've tested a few small snippets, it seems to work:

```
>> enum E { Yes; No; }
>> var a = Yes;
>> trace(a);
IhxProgram_0512.hx:14: Yes
>> trace(E)
IhxProgram_0512.hx:16: { __constructs__ => [Yes,No], prototype => { __serialize => #function:0, __enum__ => ..., __string => #function:0 }, No => No, Yes => Yes, __ename__ => [E] }
```

jcward:
**On topic:** my reaction is "meh" -- I see @back2dos's point. A few lines of class / static / import isn't a barrier to adoption. As a "learning tool", this won't move the needle (e.g. as oposed to [try.haxe.org](try.haxe.org) which is a great learning / example-sharing tool.)

And for scripting -- I'll just say it: Haxe isn't really convenient enough that I've ever wanted to write a quick script in it. It just doesn't hold a candle to the convenience of e.g. Ruby. Haxe has very different goals and ideals than most scripting languages, and that's fine.

But meh. I don't have any real arguments against the proposal.

fullofcaffeine:
> It just doesn't hold a candle to the convenience of e.g. Ruby. Haxe has
very different goals and ideals than most scripting languages, and that's
fine

Hmm... Haxe is certainly more convenient than Javascript in many regards,
and still, Javascript is considered to be a scripting language. Would you
mind to elaborate on this statement?

Justinfront:
Would allowing Main.hx to have the hxml commands above package; feel more scripty especially if there was a haxelib to generate a range of these single file templates along with the function changes. It's nice to have a single file for scripts and having a compile.hxml and Main.hx separately is not so ideal. I can't see that it would not be possible to allow/implement a mixed Main file?  But doubt if anyone would think it is a good idea beyond me :)  sorry off topic.

Pauan:
@hughsando Linux only allows for a single argument in the shebang, so that won't work.

[Nix works around that limitation by doing this:](https://nixos.org/nix/manual/#ssec-nix-shell-shebang)

```bash
#! /usr/bin/env nix-shell
#! nix-shell -i real-interpreter -p packages
```

The second `#!` line describes the actual parameters. Haxe could do something similar.

Nix already supports both Haxe and haxelib, so if Haxe ignored all lines starting with `#!` then it would be possible to write Haxe shell scripts in a single file using Nix + [haxex](https://gist.github.com/jasononeil/5432759):

```haxe
#! /usr/bin/env nix-shell
#! nix-shell -i ./haxex.sh -p haxe

class Main {
    static function main() {
        trace("Hello world!");
    }
}
```

But this is starting to get very off-topic, I think that should be in a separate proposal or issue.

jcward:
> Hmm... Haxe is certainly more convenient than Javascript in many regards,
and still, Javascript is considered to be a scripting language. Would you
mind to elaborate on this statement?

Sure. My types of "scripting" tasks are typically shell scripts. Running external commands and munging (aka processing) text files.

So I imagined myself a random scripting task -- count all the "x" characters in all the `.hx` files in the haxe repo. I timed myself, and literally took less than a minute to do this in `irb`:

```
> sum = 0; `find . -name "*.hx"`.split("\n").each { |f| sum += File.read(f).split("x").length-1 }; sum
30022
```

Now, there're all kinds of questionable practices there. And it may look like Greek to many, but the process of constructing such a program [in a REPL, with Ruby language features] is _insanely_ fast. I answered my question in less than a minute. That's what a quick scripting language does for me. Haxe -- I can't imagine how long or how many lines of code it'd take me to do this (in fairness, partially because I don't do this type of thing in Haxe.)

But I use Haxe to build large, reusable systems. Without compile-time type safety, Ruby and JS are a nightmare at that. They're different tools. I use both. It would take massive syntax / paradigm changes for Haxe to beat the convenience of Ruby at these kinds of scripting tasks. Skipping `class` boilerplate is barely a drop in the bucket. Meh.

---

Ok, for comparison's sake, I coded the above in Haxe. Here's what I ended up with: https://hastebin.com/zeyudaruci.cs

Ignoring the time I spent looking up various APIs (which wasn't really Haxe's fault), there are still a bunch of minor inconveniences that all add up:
- I had to create a new directory and Main.hx file
- I had to write some class boilerplate
- The sys.io.Process cmd/args syntax is less convenient, and I had to check the arg quotation.
- The p.exitCode() line in explicably hung for tests with a lot of files to scan... hmm. wasn't sure if I needed to read the exit code to ensure stdout was ready, saw it in some exmaple...
- lot of API calls to get from stdout -> string
- Haxe array doesn't have `each` without some static extension -- stringing together expressions isn't the norm. Instead more intermediate vars are created.
- Haxe's `String.split` interprets trailing delimiter as (empty string on the right) while Ruby's doesn't. This adds a blank filename and makes Haxe require the file exists check (or a pop would do.) Well played, Ruby.
- The dev cycle (make changes / leave editor / compile / test, repeat) was more clunky than the Ruby REPL.

hughsando:
I think most of these problems come from the process api - and maybe that is what makes a good scripting library.  But maybe not. I think its just a library thing.  Ideally you would do something like this:
```haxe
var sum=0;
FileSystem.findMatching(srcDir, "*.hx", fn -> sum+= File.getContent(fn).split("x").length-1);
println(sum);
```
Which is only one utility function (findMatching) away from being  1:1 code match.
The other thing that makes it less scripty is FileSystem, File and println are not top-level, and the boilerplate code already mentioned, which I guess we are looking to help here.

On the plus side, the haxe will work on windows without cygwin, and this is why I use haxe for my scripting, and write the utility functions as required.  eg https://github.com/haxenme/nme/blob/master/tools/make_classes/MakeClasses.hx#L7 is pretty similar.

back2dos:
Ok, I took a moment to think about it and I apologize for the length of my post. Then again I hope it contains all I have left to say on the matter ;)

First of all, "X is an interesting point but should be put in a different proposal" just doesn't work. Take the optionality of the package statement I brought up. You can consider that a different question, but there's no wisdom in doing so. The basic question is the same - "what is the guiding principle?" - and the possibility to omit the wrapping class or the package statement are part of the same answer. If you make that two questions and answer them differently, you will violate the principle of the least surprise. This is a no-go, *especially* if your declared goal is to make the language more palatable for newcomers. 

Good design is driven by principles, not by features. We're headed the opposite way here. So look: if after 12 or so years we wish to reconsider/reformulate/realign Haxe's principles, I very much welcome that. But that's not at all what is happening.

Beyond the risk of inconsistency, there is also the risk of self-incapacitation. One can pretend that the idea to make Haxe "syntactically more suitable for scripting" is entirely separate. I don't see that. It is only a separate question once you can guarantee separation, which requires a concrete solution for having this idea and the other at the same time. That may be disambiguation per file extension or what not. But we need to explore this problem at least to the point of understanding how much the cost of adding the second will increase by adding the first and **count that towards the price of adding the first**. Language features do not exist in a vacuum. Some of my favorites:

- we can't have unary postfix `?` because ternary `?` was such an existential time-saver that it had to be included (despite the fact that `if/else` does the same)
- we can't have operator overloading through static extension, because it seemed so clever to apply static extensions across implicit casts

It is pure hubris to think that you can simply add things to a design because for the moment it seems cheap and you can't think of future problems down the road just yet. Perhaps a little retrospective by Tony Hoare makes that point better than I ever can:

> I call it my billion-dollar mistake. It was the invention of the null reference in 1965. At that time, I was designing the first comprehensive type system for references in an object oriented language (ALGOL W). My goal was to ensure that all use of references should be absolutely safe, with checking performed automatically by the compiler. **But I couldn't resist the temptation to put in a null reference, simply because it was so easy to implement.** This has led to innumerable errors, vulnerabilities, and system crashes, which have probably caused a billion dollars of pain and damage in the last forty years.

Yes, of course you cannot predict all future problems. But you should really make an effort instead of simply resorting to "this is not part of the proposal". And even if you cannot think of anything, it is always more prudent to assume that [adding complexity has an inherent cost](https://m.signalvnoise.com/case-study-how-complexity-creeps-in-cba48023e6a1), so if you can't make a strong case for the benefits (and I don't see it here) that passes a compelling threshold, the arguments simply do not weigh in your favor. The burden of proof should be shouldered by those defending the proposal, not those challenging it. 

FWIW, the more I think about the proposal, the more I dislike the implementation details, because it's built on a pretty leaky abstraction. You have this implicit class and you need to take care that it doesn't pop up everywhere. It should not be in the `className` of `PosInfos`, it should not be in autocompletion, it should not be in the -xml output or dox will have to treat it specially, which leads to the broader question how haxe.rtti will actually treat it. Also: I would really want the ability to put a single function into a single module and name the module the same as the function (as is common practice with node modules, e.g. [minimist](https://github.com/substack/minimist#example)), rather than having to invent a different name for the module for no reason other than the abstraction leaking out implementation details.

I think that can be remedied. My advise would be to support it properly instead of resorting to such hackery: have top level vars and functions in all of the type system (not just rtti and documentation and macros which also need to treat `KModuleStatics`) and let the generators deal with it. In most of the cases, there's a corresponding construct, so why the indirection? Yeah, it'll be a bit more work but if it's not worth going the extra mile than I really doubt it's worth adding in the first place.

As far as reflection is concerned, I would add that there should be something like `Type.resolveGlobal` to get a top-level function/variable.

@Pauan Look. My personal opinion is that Haxe's syntax is quite baroque and could benefit from conciseness. Here is what I do: I get over it. And yes, it's my genuine opinion that people who give such syntactic trivialities enough weight to pick one language vs another have no shortage of either of the concrete mental qualities I ascribed to them. Obsessing over syntax is a guilty pleasure that I sometimes also find hard to resist and while I sympathize I do not even in the slightest consider it reasonable and I don't see that I have to. Nor do I see that anything becomes better by long-winded posts about how writing a little less code is such a huge benefit (see the irony?). I called no-one a moron and I don't insult people who disagree with me. To make such claims is baseless (I don't disagree with your preference), slanderous and intellectually lazy. Please refrain from doing that and we'll be just fine. Thank you. 

We share the same preference. That doesn't mean I will blindly accept flimsy, speculative or even quite fallacious arguments just because they support my point of view. If all that we're left with is "I like it better this way" then that's what we have and trying to conceal it behind a grand discourse is just dishonest. I'm sure you can do better than that and I invite you to do so ;)

----

I absolutely agree with Hugh that Haxe makes for a great scripting language and after overcoming some initial pain it has become my weapon of choice. The speed of Haxe's static analysis makes its cost negligible for scripting tasks and the runtime speed of eval is just fine (I'm pretty optimistic it beats Ruby for CPU-intensive stuff). Pursuing this is a worthwhile goal and would entail:

1. Having a "scripting mode" without the boilerplate.
2. Writing the utility functions that typical scripting tasks require and quite possibly autoimport them in "scripting mode".
3. Properly advertising Haxe's strength as a scripting language.

With the rise of devops and automation, scripting is no longer just a matter of quickly hacking something together in an interactive shell that you fire and forget, but by growing robust and maintainable processes and I believe Haxe could meet the increasing demand really well. We'd have to want it and then to communicate it. I believe we should, because it is something that will be relevant to existing Haxers (some of whom still use gulp or maven or whatnot) and can truly drive further adoption. 

The first step - and I cannot stress this enough - should therefore not be made harder by this proposal. With that provided, I don't care much about what happens to this, although I believe that a less hacky approach will pay off in the long run.

Aurel300:
Thanks for the post @back2dos .

I was originally mostly supportive of the proposal, but now I think the time "saved" is better spent by simply having e.g. a script which generates boilerplate code (an approach I take for my Haxe scripting), or an IDE which does the same.

As for utility functions which would be more "accessible" by being module-level, static extension is fine, or a short, one-letter module name if necessary.

As for the entire scripting tangent … While I do agree the standard Haxe API is lacking in some respects and inconsistent in many cases, this is a job for a specific library. It could provide simple functional programming for file glob patterns, directories, etc etc – whatever is generally used in scripting / batch automation.

fullofcaffeine:
Any news here? 🙄

nadako:
There's a bit of stalemate here right now. I still think that module-level functions would be a nice addition to the language without introducing a lot of new concepts to the user. At the same time, after what @back2dos said and what we discussed on Slack, I'm thinking that if we want to have module-level functions/vars, maybe we should indeed introduce them as proper type-system citizens instead of having a wrapper class auto-generated. But that's a lot more intrusive change wrt implementation, so I'm not sure if it's really worth it...

hughsando:
No point over-complicating the implementation.  From the c++ backend point of view, the auto-gen class would work well.  Any concerns w.r.t. reflection can be unit tested relatively easily and implemented gradually, which makes things easier.  If you can't tell from the outside what implementation is used, then you can't care about it.

nadako:
@hughsando I think the concern here is about macro representation mostly. Anyway there are still three options and I'm personally fine with either:

 * type module-levels in a primary module class, thus forbidding explicit primary module class - probably the simpliest approach
 * create a special class for module-levels and allow primary module class - some more work needed wrt resolution rules
 * have module-levels as a first-class citizen (so a new variant to `module_type`/`ModuleType`) - the most intrusive, but probably the most correct implementation

hughsando:
Yeah, the macro thing requires some thought.
There could always be a default implementation that bundles the module items into a default class immediately prior to the backend, so you don't need to change the backends, no matter what the internal representation is.

ncannasse:
I think that autogen class by default is good, and then because theses
classes have a specific cl_kind the platform can choose to remove the class
wrapper if it supports top level functions.

fullofcaffeine:
So do we all agree on how to proceed here? I just don't want the PR to be `open` for another year, which pretty much kills any momentum this proposal (still) has :)

nadako:
> open for another year

Hey this PR is only a couple months old :)

Anyway, I'll look into actual implementation.

fullofcaffeine:
@nadako Almost half a year :)

Anyway, no pressure whatsoever, we're an open-source community after all and it's all voluntary work. I just pinged to know the status. Btw, shouldn't we merge this? We got enough votes from core-team members no?

Cheers!

nadako:
Most people like the idea, while the actual implementation regarding macro structures is still kind of in question, yet it's described in the proposal, so I'm unsure about merging it in its current state.

jgranick:
Could we allow lowercase filenames? `openfl.net.navigateToURL` could be a reality on Haxe? 😓 

Perhaps lowercase filenames could be functional, and uppercase could be modular (though I'm generally in favor of no special magic based on file cases)

nc-x:
Is this on the roadmap for haxe 4? IMO it would be a great addition.

fullofcaffeine:
Will this ever be accepted and merged? 🤔

Simn:
https://github.com/HaxeFoundation/haxe/issues/7452

farteryhr:
btw I suggest a new class kind `module` (prefixing every field with `static`)?
I'm still with `import X.*`.
Implicit extraction without `.*` is a ton of sugar but not sure if it's too sweet that might make trouble...

PR #26, opened 2017-08-06
Inline Markup Literals

back2dos:
I actually wanted to brood about this a little while longer, but @mrcdk's initiative kinda forced my hand to present an alternative now.

Rendered version [here](https://github.com/back2dos/haxe-evolution/blob/master/proposals/0000-inline-markup.md).

### DIFF
--- a/proposals/0000-inline-markup.md
+++ b/proposals/0000-inline-markup.md
@@ -81,6 +81,8 @@ var openingTag = '<$tagName',
 
 parser.pos = start + openingTag.length;
 
+//below `hasNeither` and `count` are left as an excercise to the reader
+
 switch source.indexOf('/>', parser.pos) {
   case v if (v != -1 && hasNeither(source, parser.pos, v, ['<', '>'])):
     parser.pos = v + 2;
@@ -90,7 +92,7 @@ switch source.indexOf('/>', parser.pos) {
         case -1: throw 'unclosed $openingTag';
         case v: parser.pos = v + closingTag.length;
       }
-    } while (count(openingTag, source, start, parser.pos) != count(closingTag, source, start, pos));
+    } while (count(openingTag, source, start, parser.pos) != count(closingTag, source, start, pos)); //
 }
 
 return makeInlineMarkupExpression(source, start, parser.pos);//just turn it into some opaque expression
@@ -124,6 +126,8 @@ This feature does not impact existing non-macro code directly. Depending on how 
    ![](https://raw.githubusercontent.com/back2dos/js-virtual-dom/master/hxx-example.gif)
    
    However trying to get the parser to be reentrant (so you can write e.g. `<div>{[1,2,3].map(i -> <button>{i}</button>)}</div>`), requires you to parse Haxe syntax by hand, which not only is slow and desparate and silly, but also kills autocompletion ... unless you want to implement that part too, which is way beyond desparate and silly.
+   
+3. The idea was proposed to make such a language extension rely on compiler plugins. I'm not even sure it is possible to swap out the parser in this manner, but I would argue it is undesireable, because it makes all libraries that rely on custom parsers mutually exclusive. This is not so with macros, because macros are called explicitly through specific entry points and can therefore coexist in the same build.
 
 ## Opening possibilities
 
--- a/proposals/0000-inline-markup.md
+++ b/proposals/0000-inline-markup.md
@@ -138,4 +138,4 @@ It has been pointed out that [Haxe could greatly benefit from an easy way to emb
 
 ## Unresolved questions
 
-None so far.
+An important question is how to go about syntax highlighting within inline markup, but it seems to me that it doesn't concern the language as much as it concerns IDEs.
--- a/proposals/0000-inline-markup.md
+++ b/proposals/0000-inline-markup.md
@@ -1,0 +1,137 @@
+# Inline Markup
+
+* Proposal: [HXP-NNNN](NNNN-filename.md)
+* Author: [Juraj Kirchheim](https://github.com/back2dos)
+
+## Introduction
+
+This is an attempt to salvage the rejected [Inline XML proposal](https://github.com/HaxeFoundation/haxe-evolution/pull/12).
+The idea is to allow the compiler to identify "inline markup literals" as *opaque* expressions that can be processed by macros, with "markup" actually being as vague as possible. To illustrate what consitutes such inline markup:
+
+1. For those who want JSX:
+
+ ```haxe
+ funtion doThatReactThing() 
+   return <div class="container">
+     <div class={whatever}>
+       <a href={someUrl}>Click me!</a>
+     </div>
+   </div>
+ ```
+2. HEREDOC-like syntax has been requested for a long time now, and inline markup would allow to embed arbitrarily complex strings into haxe:
+
+ ```haxe
+ var myComplexString = 
+   <SomeDelimiter>
+     In here I can write "anything" without the parser really interfering
+   </SomeDelimiter>
+ ```
+3. Or you know, just go crazy:
+
+ ```haxe
+ function madness()
+   <JavaScript>
+     console.log("This is Spartaaaaaaa!!!");
+   </JavaScript>
+ ```
+
+4. And while you're at it, mix it any way you want it:
+
+ ```haxe
+ function someHTML() 
+   return <html lang=en>
+     <head>
+       <title>Example</title>
+       <meta name="example" content="This is just to point out that void elements are fine">
+     </head>
+     <body>
+       <style>
+         body {
+           background: #FF00FF;/* everybody loves this color */
+         }
+       </style>
+       <h1>Hello, world!<!-- hoho!!! --></h1>
+       <div contenteditable>Edit me!</div>
+       <script>
+         alert("You're joking, right?");
+       </script>
+     </body>
+   </html>
+ ```
+ 
+In each of those cases, Haxe is blisfully unaware of any language carnage going on *within* inline markup. It knows only enough to identify runs of inline markup, so that macros can pick them up and parse them. If an inline markup literal is not processed by any macro and makes it all the way to the typer, an error is generated (could just be "Unprocessed inline markup").
+
+## Motivation
+
+The main motivation here is really to make Haxe more appealing to web and desktop/mobile application developers. For better or worse, XML-ish dialects have establised themselves as the de-facto lingua franca of describing UIs for both web and native environments. XIB/Storyboard, XAML, MXML, FXML, JSX, HTML, HaxeUI markup, Stablex markup, ... you name it. If we want Haxe to gain traction outside game development, working on tighter integration between UI markup and the rest of the language is not a "nice to have" but of pivotal importance. 
+
+That said this proposal aims to be a basis that opens many possibilities instead of comitting Haxe to support any single hype, thus giving macro authors the freedom of choosing whichever sinking ship they want to put their bets on.
+
+## Detailed design
+
+If wherever an expression is expected to begin, the character `<` is found followed *directly* (i.e. no whitespace inbetween) by a letter, it signifies the start of an inline markup expressions. Then the opening "tag" is determined (this will be something of the form `<foo.Bar_barf3-gnieh:blargh`) and we read forward until we find a matching closing "tag" (which would then be `</foo.Bar_barf3-gnieh:blargh>`), while continuing the search if that tag was not ballanced. Alternatively, if there is no `<` and `>` between the opening tag and the next `/>`, then we extract that text run as a markup literal. In Haxe code:
+
+```haxe
+var start = parser.pos;//which is at this point the position of the left angular bracket preceeding the letter
+var parseTagName = ~/^[-._:a-zA-Z0-9]+/;
+parseTagName.matchSub(source, start + 1);
+var tagName = parseTagName.matched(0);
+var openingTag = '<$tagName',
+    closingTag = '</$tagName>';
+
+parser.pos = start + openingTag.length;
+
+switch source.indexOf('/>', parser.pos) {
+  case v if (v != -1 && hasNeither(source, parser.pos, v, ['<', '>'])):
+    parser.pos = v + 2;
+  default:
+    do {
+      switch source.indexOf(closingTag, parser.pos) {
+        case -1: throw 'unclosed $openingTag';
+        case v: parser.pos = v + closingTag.length;
+      }
+    } while (count(openingTag, source, start, parser.pos) != count(closingTag, source, start, pos));
+}
+
+return makeInlineMarkupExpression(source, start, parser.pos);//just turn it into some opaque expression
+```
+
+The result here would either be an `EConst(CMarkup(theMarkup))` or a `EMeta(':markup', EConst(CString(theMarkup)))` if we wish to avoid breaking changes to the AST. To repeat the above: **unprocessed markup literals should be rejected by the typer**. 
+
+It has been suggested to have a default meaning, but I maintain for now that this is practically impossible to find one that will work well for all use cases and even if it is possible, it can be put forward in a follow up proposal that can be implemented at a later time.
+
+## Impact on existing code
+
+This feature does not impact existing non-macro code directly. Depending on how it is encoded as an expresion, it may break macros.
+
+## Drawbacks
+
+1. the proposal may be somewhat modest, because it still means that people have to write macros to process the inline markup. 
+2. cases can be constructed with CDATA and comments that violate the principle of the least surprise:
+
+    ```haxe
+    var xml = <foo><![CDATA[</foo>]]></foo>;//will result in "expected ]"
+    var xml = <foo><!-- </foo> --></foo>;//will result in "unexpected >"
+    ```
+    I'm going to go with: true, these cases exist, but they don't matter. There are countless ways to deal with the issue should it arise.
+
+## Alternatives
+
+1. Parse external files at macro time. This is pretty neat, but one if the biggest problems here is that you just can't convince the compiler to give you autocompletion. And you can't blame it, because it would have to type the whole codebase to figure out which macro is actually causing the file to be parsed and call it accordingly. 
+
+2. Parse inline strings. This works *amazingly well* in `tink_hxx` (I am biased here, but I think it really shows the benefits of a proper, tight integration):
+
+   ![](https://raw.githubusercontent.com/back2dos/js-virtual-dom/master/hxx-example.gif)
+   
+   However trying to get the parser to be reentrant (so you can write e.g. `<div>{[1,2,3].map(i -> <button>{i}</button>)}</div>`), requires you to parse Haxe syntax by hand, which not only is slow and desparate and silly, but also kills autocompletion ... unless you want to implement that part too, which is way beyond desparate and silly.
+
+## Opening possibilities
+
+It has been pointed out that [Haxe could greatly benefit from an easy way to embed *any* (domain specific) language](https://github.com/HaxeFoundation/haxe-evolution/pull/12#issuecomment-306733033). This proposal happens to do that. It is fair to ask whether the surrounding delimiters need to be "tags". They don't: it could be any ASCII-art. However using tags has two advantages:
+
+1. it just so happens that most people are used to tags acting as delimiters, so we don't have to reinvent the wheel here
+2. if the embedded language is XML-ish, then the delimiter fuses with the language, which reduces visual clutter
+
+## Unresolved questions
+
+None so far.


nadako:
I like this in general, but I'm not sure if it's possible to implement such parsing in a straightforward way, given Haxe's current parser. The inline markup block should ideally come as a single token from a lexer, but it has no idea about what parser expects and whether it's at the "start of the expression", so when it finds `<tag`, it just emits two tokens (`<` and `tag` identifier). Maybe the parser could toggle some tag-lexing mode, I don't know, have to look into it.

ibilon:
An issue is that a tag can be valid haxe code.

https://try.haxe.org/#Ca85f
It's a specifically constructed example, but could happen in real code.

RealyUniqueName:
Maybe some other chars could be used instead of `<` and `>`?

benmerckx:
> An issue is that a tag can be valid haxe code.
> \<b\> isn't a tag

The proposal states:

> If wherever an expression is expected to begin, the character < is found followed directly (i.e. no whitespace inbetween) by a letter, it signifies the start of an inline markup expressions.

Which means `a<b>c;` should parse as before because `<b>` is preceded by `a` (after `a`, which is an expression, the parser expects valid cases which may come after an expression such as `a.b`, `a<b` but not the start of a new expression).

ibilon:
What about `a<<b>c` with `<<` overloaded? `a<(<b>c)` or `(a << b) > c`
IIRC `<<` is already annoying for the lexer (parser?)

But yeah we could just use some other characters, the most likely easiest (I guess) would be to introduce a new one like `#` or `\``, that way it can't be currently valid haxe code.

benmerckx:
The `<<` token will be [tokenized](https://github.com/HaxeFoundation/haxe/blob/development/src/syntax/lexer.ml#L307) before `<`. The current rules wouldn't change. Which means if you would want to compare one of these markup literals you'd have to write `a < <b></b>`, where `a<<b></b>` would understandably result in an error. While I can't be sure there's no edge case at all, I think this proposal is of little use if other characters are chosen.

nadako:
@ibilon I think it's the `>>` which is annoying because of type params.

Anyway, one of the easy ways would be to require a special character to make lexer enter tag parsing mode, not unlike how it's done for strings. E.g. ``var tag = `<div></div>`` (we don't need a closing backtick here because we apply the tag handling rules described in the proposal).

RealyUniqueName:
I'd prefer something like `@<tag>...`

back2dos:
@ibilon to build on what the others said: yes, `a<<b>>c` could be interpreted as `a << b >> c` or `a < <b> > c`, but always has and always will parse as the former, because Haxe always chooses the longest valid token. Examples:

- `1.e8` parses as a float literal, not as accessing the field `e8` on `1` ... if you want that, [you need to write `1 .e8`](https://try.haxe.org/#85432). 
- `a != b` and `a ! = b` [mean different things](https://try.haxe.org/#afc2b) just as well, where the latter is unary postfix `!` and assignment.

Matters are relatively simple (conceptually - as for implementation, that seems to be a different matter): only when the parser begins parsing a new expression does it allow for a tag to occur and there is no room for ambiguity there, because `<` is syntactically not allowed in that place.

@nadako I've tried to get some understanding of how the parser works and found very little that would make me hopeful. The one solution I see is that the parser may be able to set some shared mutable state to tell the lexer to expect a markup literal, but peeking ahead would defeat this hack entirely: `if (foo) bar; <div />` and we get "unexpected <".

So I see three options:

1. move forward with https://github.com/HaxeFoundation/haxe/issues/6477 and rely on said hack, which is this filthy as hell but would actually work (right?)
2. lex `<` followed directly by a word character as inline markup literal. It's not what I originally proposed. However it is consistent with giving longer tokens preference, but it does break a specific class of unformated code ... we already have a [migration tool](https://github.com/vshaxe/haxe-formatter) for that though ;) 
3. use the backtick.

I think the first approach is a disaster waiting to happen, but you tell me. The third solution is the most pragmatic and I could live with it. No point in obsessing over one character. I'm not really looking forward to having to explain for the tenth time why the backtick is there, but it's a lot better than nothing and we can still look into getting rid of it in the future.

That said, I'm actually intrigued by the second one (I have never seriously considered it until this point). There's the issue of having `if (a <b)` leading straight to the confusing "unterminated inline markup" but I think this could be prevented by allowing unterminated markup in the lexer (add a bool flag of course to discern them). When parsing, unterminated markup as the first token in `expr` would in fact raise "unterminated inline markup" while in `expr_next` both terminated and unterminated markup would result in "ambiguous < must be followed by whitespace" (feel free to suggest a more easily understood way of telling that). It's not perfect, but then again [what is](https://try.haxe.org/#93A0C). I'll let you guys think about that ;)

back2dos:
But yes, why not `@`. Don't see much of a difference there. Would leave backtick free for other things.

frabbit:
Not directly related, but i actually have a (working) prototype which allows an alternative calling syntax for macros via metadata like `@-myMacro expr` or `@-myMacro(config) expr`. The `-` is required to distinguish regular metadata from macro calls. That idea also allows block syntax like:

`@-myMacro {
   doSomething();
}`

which gets translated to

`myMacro({ doSomething })`

or

`@-myMacro(a,b,c) { 
doSomething
}`

which gets translated to something like
`myMacro({ doSomething }, [a,b,c]);`

ncannasse:
@nadako remarks are quite valid, you need to detail how this integrate with existing haxe lexer. Appending @ does not seem very elegant.

Actually we can detect an XML start in parse_expr by matching `<` (ident) (ident `=` const(string))* `>` in parse_expr, but then we need to switch the lexing in "raw mode" until we have close all idents. That's again feasible, but would not that be a problem for IDEs ?

For the record I found both Scala and VB have XML literals.

Also, it might be misleading how allowing what looks like "XML literals" without actually enforcing XML (or HTML) syntax. I can see how one would complain about the compiler not giving error when writing the following:

`var x = <xml> <oops/ </xml>`

So maybe using xml tags with this is a bad idea after all.

back2dos:
> you need to detail how this integrate with existing haxe lexer. 

I don't have the slightest idea how the lexer and parser work. I can whip up an implementation for hscript if that's of any help ^^

> Appending @ does not seem very elegant.

Agreed.

> Actually we can detect an XML start in parse_expr by matching < (ident) (ident = const(string))* > in parse_expr, but then we need to switch the lexing in "raw mode" until we have close all idents. 

If there is any such thing as a "raw mode", then that's the simplest solution. 

Still I have to stress that an opening tag is nothing but `<(tagname)`. See the proposal for what constitutes a tag name, because `svg:rect` is valid in XHTML, `custom-component` is valid in HTML5, `namespace.Component` is valid in JSX. The parser should not even attempt to parse attributes, because again, `<div contenteditable aria-labelledby="someLabel">Awesome content!!!</div>` is perfectly valid HTML. The syntax proposed is chosen very carefully to cover all that ground.

> That's again feasible, but would not that be a problem for IDEs ?

That depends on what "problem for IDEs" means. Getting completion to work is quite possible, as the little gif shows. Syntax highlighting is more of a challenge, as already mentioned [at the end](https://github.com/back2dos/haxe-evolution/blob/master/proposals/0000-inline-markup.md#unresolved-questions). It depends very strongly on how the IDE approaches the matter. I think an approach to highlighting that is as broad as this proposal may have a fair chance of providing a decent result. Then again I'm no expert on the matter. Perhaps @nadako can comment.

> I can see how one would complain about the compiler not giving error when writing the following:
> 
> ```
> var x = <xml> <oops/ </xml>
> ```

And how would that ever happen? According to the proposal, the above *will* give a compiler error unless a macro processes it. If the macro expects XML syntax, it will produce an error (the most naive one will just produce the exception thrown by `Xml.parse`). But if it can make sense of it, then I don't see the problem. Alter it slightly and you may have a use case:

```
var x = <html><script> console.log(5 <oops/ 4 )</script></html>
```

There's a lot of super hairy stuff in actual HTML and XML. So if then you go to having half an XML or HTML parser (which one would you choose?), you're going to ingrain incompatibilities with the actual standard *into the syntax*.

The most pragmatic approach is thus to make it the user's problem. If they want all bells and whistles, they use something like [dom4](https://github.com/therealglazou/dom4). Or whatever. And when parsing speed becomes a problem (and eval offsets that threshold quite a bit), my understanding is that they can write the parser in ocaml and just plug it in and call it from a macro (right?).

ncannasse:
Actually I'm taking back my previous comment. We could ensure that when getting compiled (without going through macros), XML literals are checked against XML by the compiler and parsed in a similar way to Haxe smart strings. This way we get both something that doesn't require macros and can be used by macros in a smart manner.

I still have one particular issue: your proposal leaves attributes syntax undefined, but it means the following would also be invalid syntax:

`var s = <foo>${if (x<foo) 0 else 1}</foo>` because of the `<foo` being mistaken as a opening tag, leading to quite hard to debug error messages about the first `<foo>` not being closed.

Maybe we should ensure some XML attribute syntax in order to distinguish an actual node from something else.

kevinresol:
I personally prefer not to include the xml parsing part into the compiler, mainly because of the looooooong release cycle. 

I found myself always in a "I want to use haxe nighty but I can't use haxe nightly" dilemma. 

For example, I want to use haxe nighty for these fixes:
- https://github.com/HaxeFoundation/haxe/issues/6224
- https://github.com/HaxeFoundation/haxe/issues/6539
- https://github.com/HaxeFoundation/haxe/issues/6532

But I can't because https://github.com/HaxeFoundation/haxe/issues/6321 breaks all my codes.

Sorry for being a bit off-track here, but my point is: please don't embed functionalities into the compiler if there are other alternatives. I think the same reason drove the team to move out SPOD as a macro library.

ncannasse:
@kevinresol I would agree in general not to put too much things in compiler if it can be avoided (SPOD required some compiler specific support a long time ago - before macro era, it can be safely removed now)

However I don't think that having a syntax that requires a macro and lead to a compiler error otherwise is a good idea, especially if there's some behavior that can be expected from the end user. It seems obvious to me that the following code should compile if it's considered valid syntax: 
```var x = <foo>$str</foo>```

**Comment by szczepanpp** [comment]
@ncannasse that's one of my ideas (I've had to share with anybody yet) to use interpolation syntax as a straight-forward solution for embedded DSLs (yes, I did notice this proposal is about "markup literals" ;)). It would require generation of "interpolation AST", so too speak, that would be available during macro processing. If there were no macros transforming it, every target would have to simply generate runtime interpolation, exactly as it's done now. This way every "inline DSL" would just be a string by default. Added bonus would be an easy way to find out which actual Haxe variables are referenced inside the interpolation/DSL string. I'm omitting now the need of finding out what macro to run on any given interpolation usage. Do you think it's a viable solution to this problem?

**Comment by Simn** [comment]
I still don't really know how people expect this to be actually implemented without implementing a full XML parser. Counting opening and closing XML tags is all nice and fun until you have `<tag ... />`. Can't parse that without understanding XML syntax, and we voted against integrating a XML parser before, so I don't really understand what we are discussing here.

I'm not saying that I'm strictly against something like this, but I'm still not sure if XML is the solution.

**Comment by markknol** [comment]
Something I think that should be given some thoughts is how to deal with code comments. If the rule is `<tag ANYTHING </tag>` then this can lead to unexpected results.
```haxe
var x = <div>
     // </div>; // Am I closing tag, content or comment
        </div>;
```

Simn:
As I understand it, the idea is to leave the Haxe domain upon `<div>`, so it should no longer recognize Haxe comments.

But yes, literals in general are also a good point: In order to parse this correctly, you'd still have to understand e.g. string literals in order to not close on `<div x="</div>">`...

fzzr-:
IMHO. I don't like to mix markup lang in the any lang like Haxe. But i think it can be cool feature if it implement in isolated context like a block and builtin macro, look:

```haxe
function foo()
{
    var myXml = @markup(XML) {
        <node foo="bar"/>
    }

    var myYaml = @markup(YAML) {
        foo:
            - "bar"
            - "far"
    }
}
```

This method provides modularity and extensibility.


- - -


I'll try to explain my strange idea :)

For this method we need one none-breaking compiler modification - new type of `Expr` - `UnsafeBlockExpr`.
So,

- All expressions in the `{` block `}` marked with registered `meta` is a `UnsafeBlockExpr`
- A build-function (like for `@build`) should receive two arguments:
    * meta arguments - `@markup(THIS_ARG, AND, THIS)`,
    * arg `:UnsafeBlockExpr` and return `:Expr`

_What is that `UnsafeBlockExpr` and for what?_
Any "raving lunatic" expressions in this block should be ignored by the compiler until registered _MarkupBuilder_-function returns `Expr` not `null`.
Else or if no suitable registered _MarkupBuilder_-function was than the compiler should generate an error about "have no suitable Markup Builder macro func...".


And I can say with confidence what any markup in future will can be implemented on `macro` only without modifying the compiler.


And abstract example:

```haxe
class Main
{
    static function main()
    {
        var myValue = 42;
        var myXml = @markup(XML) {
            <?xml version="1.0" encoding="utf-8"?>
            <x>
                <ml value=$myValue/>
            </x>
        }
        $type(myXml); // Xml

        var myYaml = @markup(YAML) {
            foo:
                - "bar"
                - ${myValue}
        }
    }
}

// very stupid simple example of custom builder
class CustomMarkupBuilder
{
    public static macro function init()
    {
        // similarly to Compiler.addGlobalMetadata
        Compiler.setCustomMarkupGenerator(":markup", "XML", buildXml)
        Compiler.setCustomMarkupGenerator(":markup", "YAML", buildYaml)
    }

    public static macro function buildXml(srcExpr:UnsafeBlockExpr):Expr
    {
        var src = srcExpr.toString();
        var xml = Xml.parse(src);
        return macro Xml.parse(${srcExpr});
    }

    public static macro function buildYaml(srcExpr:UnsafeBlockExpr):Expr
        return macro null;
}
```

and call `init` as initial macro, frag of build.hxml:

```hxml
--macro "CustomMarkupBuilder.init()"
```

- - -

This is only my hopes and dreams. But I'm sure it would be very cool!

fullofcaffeine:
Any news here?

EricBishton:
This type of markup from the Nemerle programming language (http://nemerle.org/About) looks very clean:

```Nemerle
def title = "Programming language authors";
def authors = ["Anders Hejlsberg", "Simon Peyton-Jones"];
    
// 'xml' - macro from Nemerle.Xml.Macro library which alows to inline XML literals into the nemerle-code
def html = xml <#
  <html>
    <head>
      <title>$title</title>
    </head>
    <body>
      <ul $when(authors.Any())>
        <li $foreach(author in authors)>$author</li>
      </ul>
    </body>
  </html>
#>
Trace.Assert(html.GetType().Equals(typeof(XElement)));
WriteLine(html.GetType());
```

I rather like a few aspects of this:  First, the operators '<#' and '#>' as markup(?) delimiters is very readable; Second, the formatting of the xml as a "here" doc (_a_ _la_ bash) makes for very clean code; Third, the keyword (rather than a @macro(something) { stuff here }) is shorter, easier to type, and makes very clear the context.  

So, more in a Haxe parlance, I would see it like this:
```haxe
import Sys;

class Test {
  static function main() {
    var title = "Programming language authors";
    var authors:Array<String> = ["Nicolas", "Simon", "<it>et al</it>"];

    var xml:String = @lang(xml <#
        <html>
          <head>
            <title>$title</title>
    	  </head>
    	  <body>
            ${if (authors.size()) @lang(xml <#
       	      <ul>
                ${for (a in authors) @lang(xml <# <li>$a</li> #>) }
              </ul>
              #> }
          </body>
        </html>
      #>);

    trace(xml);
  }
}
```
It's not quite as clean because of the nice $when and $foreach that Nemerle has, but it's still fairly nice.

What @lang() does is:
0) Determine which language plugin is to be used, and if it's missing stop the compilation (IIRC, already implemented in Haxe 4);
1) Catenate all characters between the beginning and ending tokens (nesting is allowed!);
2) Then it does interpolation on the collected string (we can disallow that with a '<!#' begin token);
3) Then, it passes the final string off to the language plugin for parsing/processing.

The big benefit with this approach is that the Haxe compiler *doesn't have to understand the embedded language*.

Can this be done with a macro already?  I don't know; it seems like it, except possibly for the token parsing.  (It would be a pretty hefty macro to do parsing and validation!)

fullofcaffeine:
Maybe if integrating this into the core compiler is not something the core team wants to do right now, what if we create an external tool to pre-process this? I have two questions/suggestions regarding that:

1) It could possibly be a frontend to the Haxe compiler. It would ignore everything else and pass it over verbatim, but would transform those templates into something Haxe could understand (something like what tink_hxx does nowadays). The only thing is that this binary would have to be called instead of the actual `haxe` compiler, which could be awkward to setup.

2) Could the Haxe ocaml plugin architecture be used for it? This way this would be implemented as an external project to the core language/compiler. 

I agree with @kevinresol that it probably shouldn't be part of core, mainly because of the extreme long release cycles.

EricBishton:
My thought was to use an ocaml compiler plugin.  I'm not sure if it can currently be implemented separately without any new compiler support, but that's the goal.  I would expect a number of extension projects for the various types of output.  HF would probably like to manage popular technologies (such as XML, HTML, CSS).

If you want mixed code and markup, then that will be difficult to do with just a preprocessor.  (However, if all you really want is "here docs," then that can be easily accomplished with a preprocessor, and probably is easily done with macro already).

nadako:
All these `@lang`/`@markup` don't really look useful when it comes to reentrancy tbh, a plain `jsx('...')` call looks better to me.

Maybe we should consider just supporting JSX on a language level similar to [how Reason does](https://reasonml.github.io/guide/language/jsx/) and be done with it. Also JSX is really not XML, so the default behaviour should really be just desugaring into method calls.

Long release cycle is something we can fix independently, now when @Simn improved and documented the process :)

benmerckx:
> Maybe we should consider just supporting JSX on a language level similar to how Reason does and be done with it. Also JSX is really not XML, so the default behaviour should really be just desugaring into method calls.

The way Reason handles jsx is pretty clean and well-defined. Might be a good start and would still allow to be somewhat extendable through macros, as long as attribute values or children are valid Haxe syntax.

EricBishton:
> ... a plain `jsx('...')` call looks better to me.

I agree, for the most part.  (I like the '<#', '#>' operators -- can't say why.)  I was just trying to figure out how to do it without adding a new language keyword for every type of markup that we should ignore (or forcing us to pick _just_ _one_...).  Also, we really don't want to have to add markup parsing and evaluation in the compiler.

EricBishton:
> All these ... don't really look useful when it comes to reentrancy tbh...

@nadako Are you thinking of compile-time or run-time reentrancy issues?

djaonourside:
The `UnsafeBlockExpr` solution suggested by @fzzr is realy needed thing. Haxe gurus please take your notice on it:)

kevinresol:
@EricBishton Rentrancy is a syntax problem, so it is compile time.

fullofcaffeine:
>Maybe we should consider just supporting JSX on a language level similar to how Reason does and >be done with it. Also JSX is really not XML, so the default behaviour should really be just >desugaring into method calls.

I like this idea, it's bold and focuses on keeping things straightforward and simple.

I suggest we consider this as the implementation approach for this proposal and let the core team vote on this, unless we want to wait a couple more months... anyone? :)

ncannasse:
> Maybe we should consider just supporting JSX on a language level similar to how Reason does and be done with it. Also JSX is really not XML, so the default behaviour should really be just desugaring into method calls.

I think that's very short sighted. While today JSX is the hot thing, it might be something else entirely in two years. Language design is about creating solid bridges for the future, instead of single-usage wooden ladders.

nadako:
Yeah, I agree to that in general. It's just that I don't really see how we can implement arbitrary syntaxes without pluggable parser/lexer or ugly and useless (for jsx at least) heredoc syntax.

EricBishton:
@nadako - I don't think you can.   To get _arbitrary_ syntaxes, you need to be able to _arbitrarily_ extend the parser; something has to parse it.  And there must be a way to delineate it, whether using a heredoc delimiter or a meta.  

In truth, @fzzr- and I have shown very similar solutions (his UnsafeBlock example also uses string interpolation).  From an IDE implementation point of view, having a dedicated operator and being able to treat the inserted language as a string is a simpler implementation.  To treat it as an UnsafeBlock is harder, but really only because we can't detect it directly in the lexer (like we can a new operator).  At that point, we either have to have meta support in the lexer (which is not good) or all supportable syntaxes have to be Haxe syntax compatible (which is untenable).

EricBishton:
As another thought... In truth, we don't require a meta at all.  We can create the new operator similar to a markdown tag:
```haxe
import Sys;

class Test {
  static function main() {
    var title = "Programming language authors";
    var authors:Array<String> = ["Nicolas", "Simon", "<it>et al</it>"];

    var xml:String = <#xml
        <html>
          <head>
            <title>$title</title>
    	  </head>
    	  <body>
            ${if (authors.size()) <#xml
       	      <ul>
                ${for (a in authors) <#xml <li>$a</li> #>}
              </ul>
              #>}
          </body>
        </html>
      #>;

    trace(xml);
  }
}
```
Then, all we've really done is create a new string type that allows/requires an (ocaml) compiler to parse and/or verify it.  If the plugin isn't available (or there is no name immediately following the opening operator), it's still just a string to Haxe.  And, frankly, this is all people are really asking for: an easier way to embed their markup, which often requires its own string delimiters.

(I still like '<#' and '#>' better than triple-backticks (```), tough.)

markknol:
Does anyone have an example of how a plugin that should support this should look like?

markknol:
@EricBishton The haxe parser still has to parse `<#` and find `#>`. If the content inbetween is unparsed, how can you nest other blocks in it? How does it make a difference between `<#` as Haxe code and `<#` as arbitrary syntaxe code?

fullofcaffeine:
>Yeah, I agree to that in general. It's just that I don't really see how we can implement arbitrary >syntaxes without pluggable parser/lexer or ugly and useless (for jsx at least) heredoc syntax.

I was under the impression that the new plugin system allows this... no?

EricBishton:
It is true, the Haxe lexer does have to be extended to use `<#` and `#>` as string delimiters.  They are easy to add and understand because of the way the lexer works, whereas `<div>` is ambiguous because (as was discussed above), `<` followed by letters is a default lexer word separation.

Second, the content is not _unparsed_.  It is a Haxe string, subject to normal string interpolation, therefore must(?) be nested.  (And, if we want an option for a non-interpolated block, we can add `!#`, `<!#` or some other variation.)  However, string itself is unvalidated by the Haxe parser, and would be able to be re-parsed and/or validated by a plugin.

EricBishton:
BTW, we can come up with all sorts of delimiter combinations.  Perhaps `{'` and `'}` for interpolated delimiters and `{"` and `"}`  for non-interpolated.  Those actually look like current syntax:

```haxe
    var xml:String = {'xml
        <html>
          <head>
            <title>$title</title>
    	  </head>
    	  <body>
            ${if (authors.size()) {'xml
       	      <ul>
                ${for (a in authors) {'xml <li>$a</li>'} }
              </ul>
              '} }
          </body>
        </html>
      '};
```
I kind of get lost looking for the closing `'}` operator vs. the closing `}` operator for the interpolated code.

In the end, there are two goals:
1) Easily (e.g. without requiring escape sequences) allow other markup or other language code to be embedded within Haxe;
2) Allow (NOT require) the same to be parsed and/or validated during the Haxe compilation.

To do that, we need to:
1) Choose an _unlikely_ delimiter that is unused in other _expected_ embeddable languages.
2) Have a way to specify the validation.

markknol:
> the Haxe ocaml plugin architecture

I am against a plugin system where one has to use ocaml . The whole point of Haxe is to have one language that rules them all, so that would be a weird contradiction. Also, I wonder how many are willing to learn ocaml to make such plugin. I think if everything is structured and available in the well known macro context, that would be more convenient, no?

EricBishton:
> I am against a plugin system where one has to use ocaml.

I agree wholeheartedly.  Unfortunately, that's not what is currently implemented and available; we _have_ the ocaml plugin architecture already in the compiler.  I think (though I don't know) that it would be harder to implement a parser/lexer/validator in a Haxe macro than it is in Ocaml.  

As I think @ousado said when the plugin architecture was implemented, "All we need now is an Ocaml back-end for Haxe."  My corollary is, "All we need is a Haxe (or Neko, or HL) binding for Ocaml."

djaonourside:
> While today JSX is the hot thing, it might be something else entirely in two years. 
@ncannasse IMHO You aren't right in this question. Nowdays using JSX and react-like libraries is the fastest, convenient and advanced way to develop web applications. I think it will continue to evolve. So we need more advanced supporting jsx syntax in haxe(or a mechanism that can provide it) than just macro function with a string param. This thing can help popularize haxe among  potential js-target users. It's realy big audience.

impaler:
One cool use case that is not React, would be to see haxeui and other ui libs use it :)

> While today JSX is the hot thing, it might be something else entirely in two years.

Apart from adoption and time I don't know how we could have a good measurement on whether something is just a `hot thing` or not. React has been around for about 5 years and it's adoption is quite impressive. Writing ui markup in xml like formats has obviouisly been around much longer.

As for `jsx` I don't think there have been too many changes apart from addition of the [Fragment syntax sugar](https://reactjs.org/docs/fragments.html#short-syntax) `<></>`. Also interestingly there is a project for this in php https://github.com/hhvm/xhp-lib.

haxiomic:
What about using a similar syntax to [ES6 template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_templates)

```haxe
macro function jsx(string, expressions) {
    // parse jsx, return react objects
}

function render() {
    var message = 'hello world';
    return jsx`
        <div>$message</div>
    `;
}
```

back2dos:
> It doesn't lock us into a single DSL – implementing XML or JSX can be offloaded to the community rather than the compiler team

Which part about this proposal locks us into a single DSL?

> Will be familiar to ES6 JavaScript developers and may be adopted by other languages

I'm inclined to doubt that ES6 developers who're looking for familiarity will go for Haxe when they can use FlowType or TypeScript, especially since both of them have JSX support.

----

I don't mind template literals. It's a neat feature although I'm not sure of the improvement here, beyond having a third type of quote that means you can use single and double quotes in the string. You can already do:

```haxe
function render() {
    var message = 'hello world';
    return @jsx'
        <div>$message</div>
    ';
}
```

djaonourside:
I wonder what's the status of native xml support implementation:)

djaonourside:
@kevinresol Thanks. I've just noticed and removed the question)

Simn:
https://github.com/HaxeFoundation/haxe/pull/7438

PR #28, opened 2017-09-24
Type parameter variance of enum

shohei909:
This proposes new variance rule for enum.

[Rendered version](https://github.com/shohei909/haxe-evolution/blob/proposal/variance-of-enum/proposals/0000-variance-of-enum.md)

### DIFF
--- a/proposals/0000-variance-of-enum.md
+++ b/proposals/0000-variance-of-enum.md
@@ -12,7 +12,15 @@ Supports type parameter covariance and contravariance of enum.
 A common case is to assign an `Option` to another `Option`.
 
 ```haxe
-var option:Option<Float> = Option.Some(1);
+import haxe.ds.Option;
+
+class Main {
+	static public function main() {
+		var optionFloat:Option<Float>;
+		var optionInt = Option.Some(1);
+		optionFloat = optionInt;
+	}
+}
 ```
 
 Currently, following compile errors occur.
--- a/proposals/0000-variance-of-enum.md
+++ b/proposals/0000-variance-of-enum.md
@@ -1,0 +1,112 @@
+# Type parameter variance of enum
+
+* Proposal: HXP-NNNN
+* Author: [shohei909](https://github.com/shohei909)
+
+## Introduction
+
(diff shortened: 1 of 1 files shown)


Simn:
Interesting... but I wonder, does this actually solve any problem? I can see your reasoning from a typing theory point of view, but I'm not convinced that the additional complexity actually comes with any advantage.

shohei909:
Of course, this has advantages.  I added an actual example.

Simn:
To clarify, this is the failing case:

```haxe
import haxe.ds.Option;

class Main {
	static public function main() {
		var optionFloat:Option<Float>;
		var optionInt = Option.Some(1);
		optionFloat = optionInt;
	}
}
```

Your direct assignment works already due to top-down inference.

And yes I can see the point here. Given that enum arguments are read-only, covariance could be allowed.

@ncannasse: Any opinions?

ncannasse:
The problem is that type parameters variance depends on the way it's used.
For instance:

```
enum Option<T> {
    None;
    Some( f : Void -> T );
}
```

In this case `T` is contravariant...

shohei909:
If you mean this `Option1`, yes, the `Option1.T` is contravariant.

```
enum Option1<T> {
    None;
    Some( f : T -> Void );
}
```

However, `Option1` is different type from `Option2`:


```
enum Option2<T> {
    None;
    Some( f : T );
}
```

`Option2.T` is covariant without regard to usage.

This is the list of "A is B" relation:

| A | B |
----|---- 
| `Int` | `Float` |
| `Float -> Void` | `Int -> Void` |
| `Option1<Float>` | `Option1<Int>` |
| `Option1<Int -> Void>` | `Option1<Float -> Void>` |
| `Option2<Int>` | `Option2<Float>` |
| `Option2<Float -> Void>` | `Option2<Int -> Void>` |

nadako:

We concluded that we should approach this in a broader scope, not just for enum, but also in other cases, especially with "read-only" types (e.g. https://github.com/HaxeFoundation/haxe/issues/9611).

Issue #22, opened 2017-04-30
Allow (some) keywords as field names / var names


kevinresol:
Since field access is pretty much predictable, I think it _should_ be ok.

The main problem lies in `import MyClass.*` where the field access becomes unpredictable (and this becomes the same case as var names). So keywords that can act as the beginning of an expression, such as `var`, `function`, controls keywords (e.g. `if`, `switch`, `while`...), should still be forbidden.

So I think some of the possible candidates are:

```haxe
obj.default();
obj.enum;
obj.in();
obj.macro();
```
etc...

nadako:
I guess this is a question of context-specific keywords which should be possible, and even useful, but not I'm sure what long-term consequences that's gonna have. (e.g. if we allow `enum` as an identifier, i'll be harder to introduce local enum type declaration if we'll ever want it).

kevinresol:
In that case, what if we only allow field access, so leaving `import *` and `var` as-is? In other words, `MyClass.default(...)` would work, while `import MyClass.*` + `default(...)` would result in the usual parse error.

Btw, i want to mention that among all keywords, I think the most "wanted" one would be `default` and `in`, which I encounter most in extern APIs and external data (e.g. json)

Simn:
This is not worth it. It would complicate the parser for no real benefit other than a mildly nicer to read field access in some rare cases. There's already some inconsistency with the `macro` and `extern` keywords being allowed in package names, but that doesn't mean that we have to make it worse.

I'm pretty sure this would also annoy many IDEs.

kevinresol:
My concrete proposal summary:

| Location | Keywords allowed? | Remarks |
| -- | -- | -- |
| Member function/var | Yes | Must be accessed through `this.keyword`, `inst.keyword` or `MyClass.keyword`. `import MyClass.*` will not expose the function/var`
| Local function/var | No | |
| Package name | Yes | Except that keywords cannot serve as root package, i.e. `package haxe.macro` is allowed, while `package macro.whatever` is not. |

PR #32, opened 2017-10-04
`@:native` for structure fields

nadako:
[Rendered version](https://github.com/nadako/haxe-evolution/blob/anon_native/proposals/0000-anon-native.md)

---

```haxe
typedef JsonSchema = {
  @:native("enum") var enum_:Array<JsonValue>;
}

schema.enum_ = [1,2,3]; // generates schema.enum = [1,2,3]
print(schema.enum_); // generates print(schema.enum)
```

See https://github.com/HaxeFoundation/haxe/issues/5105 and https://github.com/HaxeFoundation/haxe/pull/2185

### DIFF
--- a/proposals/0000-anon-native.md
+++ b/proposals/0000-anon-native.md
@@ -38,8 +38,8 @@ When unifying anonymous structures, we add additional check for `@:native` metad
  * do the usual kind and visibility checks
  * let `n1` and `n2` be the string values of optional `@:native` metadata of these fields
  * if both `n1` and `n2` are absent, fields unify, proceed to type checks
- * if both `n1` and `n2` present and have the same values, fields unify, proceed to type checks
- * if both `n1` and `n2` present, but have different value, fields don't unify, emit an error
+ * if both `n1` and `n2` present and have the same value, fields unify, proceed to type checks
+ * if both `n1` and `n2` present, but have different values, fields don't unify, emit an error
  * if either of `n1`, `n2` is present while another one is absent, fields don't unify, emit an error
 
 This will ensure that it's safe to pass values of different anonymous structure types with `@:native` is involved.
--- a/proposals/0000-anon-native.md
+++ b/proposals/0000-anon-native.md
@@ -75,7 +75,7 @@ This should provide a straightforward and safe way of representing these kind of
 
 At the moment, `@:native` on structure fields is not handled in any way, so this should not break any existing behaviour unless there's some particular macro processing, but given the semantics `@:native` has on classes, I don't think this meta is used for anything other than the proposed feature.
 
-I know just one [macro implementation](https://github.com/clemos/haxe-js-kit/blob/develop/util/NativeMap.hx) that uses `@:native` on structures to provide the field renaming feature by @clemos that is used like [this](https://github.com/clemos/haxe-js-kit/blob/develop/js/browser/socketio/Client.hx). With the addition of the proposed feature, the macro shouldn't break, but will become obsolete and could be removed, I think.
+I know just one [macro implementation](https://github.com/clemos/haxe-js-kit/blob/develop/util/NativeMap.hx) that uses `@:native` on structures to provide the field renaming feature by @clemos that is used like [this](https://github.com/clemos/haxe-js-kit/blob/6447e9f07a430836a7f4f5eafd2f8f71895639ea/js/atomshell/browser/BrowserWindow.hx#L15-L35). With the addition of the proposed feature, the macro shouldn't break, but will become obsolete and could be removed, I think.
 
 ## Drawbacks
 
--- a/proposals/0000-anon-native.md
+++ b/proposals/0000-anon-native.md
@@ -30,7 +30,7 @@ However it has one serious flaw: if the object contains a field name that is a r
 
 From the user's perspective it's simply about adding `@:native("fieldName")` metadata to the anonymous structure definition, as shown in the example above. So `@:native` has similar/same semantics like it does in extern classes.
 
-However care should be taken not to lose `@:native` information between structure unifications, so costraints are required in that part.
+However care should be taken not to lose `@:native` information between structure unifications, so some constraints are required in that part.
 
 When unifying anonymous structures, we add additional check for `@:native` metadata on fields with the following rules:
 
--- a/proposals/0000-anon-native.md
+++ b/proposals/0000-anon-native.md
@@ -30,7 +30,7 @@ However it has one serious flaw: if the object contains a field name that is a r
 
 From the user's perspective it's simply about adding `@:native("fieldName")` metadata to the anonymous structure definition, as shown in the example above.
 
-However care should be taken not to lose `@:native` information between structure unifications, so some design is required in that part.
+However care should be taken not to lose `@:native` information between structure unifications, so costraints are required in that part.
 
 When unifying anonymous structures, we add additional check for `@:native` metadata on fields with the following rules:
 
--- a/proposals/0000-anon-native.md
+++ b/proposals/0000-anon-native.md
@@ -16,6 +16,10 @@ schema.enum_ = [1,2,3]; // generates schema.enum = [1,2,3]
 print(schema.enum_); // generates print(schema.enum)
 ```
 
+Related issues and discussions:
+ * https://github.com/HaxeFoundation/haxe/issues/5105
+ * https://github.com/HaxeFoundation/haxe/pull/2185
+
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


nadako:
I can work on implementing this, but I need some review and approval of the idea by @ncannasse and/or @Simn first.

ncannasse:
I think an easier version would be to allow `{ "enum" : 5 }` and have it typed as `{ "enum" : Int }` (both would be allowed in terms of Haxe syntax).

In that case it might still be escaped as _enum or $enum or something else if `enum` is a keyword in the generated platform.

nadako:
> I think an easier version would be to allow { "enum" : 5 } and have it typed as { "enum" : Int } (both would be allowed in terms of Haxe syntax).

That would also require some field access syntax like `schema."enum"`, but yes that would also do the job. Althought I thought you were against exactly this in https://github.com/HaxeFoundation/haxe/pull/2185 and asked for the `@:native` proposal.

What should we do?

ncannasse:
Uhm right, I didn't took field access into account, and we don't want to
allow either  value."enum" or value["enum"]

nadako:
> we don't want to allow either  value."enum" or value["enum"]

While I can see why we don't want the general `value["enum"]` syntax (since it looks like a map lookup), I'm not sure why you're against `value."enum"` - it's clearly a field access and doesn't suggest any map-like behaviour. I consider this kind of identifier escaping.

Btw, I think C# has `value.@enum` syntax for this, which could be an option too (tho maybe can be a bit confusing wrt our metadata syntax).

nadako:
Anyway, I'm waiting directions on what to do next: proceed with the proposed `@:native` stuff or implement the quoted fields syntax.

clemos:
While I think it would be nice to be able to use `value.enum` directly (I have no idea but maybe it would be possible to differentiate field names and keywords ? In JS, using such keywords as field names is perfectly valid), I think it should be done in another feature request.

ncannasse:
@Simn what do you think about us allowing `<expr>.<keyword>` ?

Simn:
Well, let's talk about consistency here:

* For fields on types, we have to use `@:native("keyword")` metadata
* For structure values, we can use quoted fields: `{ "keyword": value }`
* For field access, we would use `expr.keyword`

That doesn't strike me as good design. Allowing `expr."keyword"` would at least be consistent with the structure values, and also at least similar to the type field version because "at least they both use string literals".

kevinresol:
I am bringing this up again: https://github.com/HaxeFoundation/haxe-evolution/issues/22.

I suppose in some other languages there are concepts of contextual keywords (e.g. `await` in C#), that some identifiers are only considered as keywords in particular contexts.

IMO, we could implement that step by step. As starter, `field.keyword` should be allowed as there is no ambiguity.

clemos:
@Simn Ultimately, I think it would make sense to allow `{ keyword : value }` too (once again, like JS does).
Fields on types certainly bring more issues, but I think we could consider allowing them too (could only be accessed via `this.keyword` or `Class.keyword` though)
Anyway, I agree with @kevinresol : #22 is probably worth reopening.

nadako:
can we go with `@:native` solution that works for now and then you guys discuss new syntax?:)

ncannasse:
I fail to see how that would even work, for instance with this:

```haxe
typedef Args = { var x : Int; @:native("new") var y : Int; };

class Test {
    static function foo() return { x : 0, y : 0 };
    static function main() {
         var v : Args = foo();
    }
}
```

If it's just to allow assigning literals I think it's not good enough.

kevinresol:
The fun fact is that on js target you can use keywords as object field:

`var o = {new:1}` this runs fine in js

clemos:
@ncannasse well `@:native` already has this problem https://try.haxe.org/#36B22 so this wouldn't really be a con for this feature :S 
Obviously, I think everybody agrees allowing keywords as field names would be cleaner...

nadako:
Seriously, people? I covered this exact case in the proposal...

kevinresol:
> Seriously, people? I covered this exact case in the proposal...

Sorry I misunderstood Nicolas's comment. Yeah, the example case make sense.

nadako:
And no, I think allowing keywords is not a solution, because you still can't use fields like `not-valid` in @clemos example. So It's either the quoted field syntax or `@:native`, the latter being easier to implement and not introducing new syntax.

kevinresol:
I think the keyword stuff is a bit off-topic here, sorry for bring that out at first place. (but I still hope it could be re-considered)
Personally I think OP is a good solution with least effort in current situation.

ncannasse:
Ok, I forgot that you have dealt with it in the proposal, sorry.

Reading again the whole thing, while I'm still not 100% satisfied with the solution, I don't have either a better proposal to make at this point, so feel free to move forward with the change.

kLabz:
Any news on this?
Would be much appreciated when dealing with externs for react libs :)

FuzzyWuzzie:
I also need this for dealing with externs for various projects, most lately: Blender externs.

Simn:
It bothers me that this affects unification... It shouldn't be necessary to have typing jump through hoops because the syntax is insufficient to express what you want to express. It's not even strictly related: The shortcoming here is in the field-access syntax, and our solution is to meddle with the struct declaration and its unification behavior.

I don't know if we're still voting, but this one would get a "meh" from me. I acknowledge the problem, but I don't think this is the best solution.

nadako:
Well, but the problem is severe and the fix like this, even temporary would be appreciated, also I don't think the change in the code base will be that intrusive. Anyway, suggestions are always welcome :)

Simn:
I still think `field."name"` is the obvious solution here. It solves exactly what we want to solve and whatever problems it brings at generation-level would also be brought by `@:native`.

FuzzyWuzzie:
I don't know anything about the compiler or unification, but I can add from a user's perspective the original proposal matches exactly what I expect to happen based on how `@:native` works in other extern places (and I was a bit surprised at first when I naively tried to apply `@:native` on a field name and it didn't work like it does elsewhere). I get that "working the same as elsewhere" might not be strictly correct from a compiler / language standpoint, but I think it meshes with the intention of the `@:native` metadata (to transform Haxe types into native types, which would be expanded into: transforming Haxe types and fields into native types and fields).

I think the `field."name"` syntax is a bit confusing—it suggests to me that I could do something like `field.'name${5*25}'`, or even `var s:String = "abc"; field.s = true;` (what would it export as? `field.s` or `field.abc`? Or throw an error?)

markknol:
I'm in favor of allowing `@:native` on typedefs. It would help on the my [hxobfuscator](https://github.com/markknol/hxobfuscator) project, which shortens/obfuscates field names by decorating fields with `@:native`.

kLabz:
It seems like we will need some sort of solution to this problem in the near future for haxe js (well, react): React will most probably drop `className` in favor of `class` in the next major version ([discussion here](https://github.com/facebook/react/issues/13525#issuecomment-417818906)) (+ `for` instead of `htmlFor`).

I'm not even sure `@:native` would be enough for this use case, since when doing `jsx('<div class="myclass">hello world</div>')` there isn't any typedef for the props atm. Well, this one needs a macro anyway so it could be handled somehow.

Simn:
> or even var s:String = "abc"; field.s = true; (what would it export as? field.s or field.abc? Or throw an error?)

How would `field.s` ever resolve to `field.abc` or throw an error? It's the normal field syntax and totally unaffected by anything.

Anyway, it feels like I'm on a different planet than everybody else here. So if you want to go ahead with `@:native` I'm not gonna stop you.

kLabz:
Yep :+1:

What about typing this with typedefs? Would it be `typedef A = { "keyword":String }` / `typedef A = { var "keyword":String; }`?

nadako:
I'm gonna close this proposal as it's clear that we don't have consensus. As I mentioned before, I think `obj."field"` is a nice(r) option as well, I just thought it was rejected before. Anyway, this will require another design proposal that someone has to make.

kLabz:
@ncannasse 
> we don't want to allow either  value."enum" or value["enum"]

Do you have a strong opinion against `value."enum"`? If so, could you elaborate a little?

nadako:
FWIW I implemented `field."name"` syntax here https://github.com/HaxeFoundation/haxe/pull/9433. Can't say I'm a huge fan of this, but we need _some_ solution.

markknol:
So this issue can only be tackled at the call side and not at the definition side?

nadako:
> So this issue can only be tackled at the call side and not at the definition side?

With `@:native` solution, it's just the definition. With field-quoting solution it's both definition and access.

markknol:
Ah that sounds great!

sonygod:
I can't wait  for this.

PR #38, opened 2017-11-17
stack based tuples

jdonaldson:
Here's a proposal for stack-based tuples.

### DIFF
--- a/proposals/0004-stack-based-tuples.md
+++ b/proposals/0004-stack-based-tuples.md
@@ -1,0 +1,148 @@
+# Stacktypes : Stack-based Tuples
+
+* Proposal: [HXP-0004](0004-stack-based-tuples.md)
+* Author: [Justin Donaldson](https://github.com/jdonaldson)
+
+## Introduction
+
+Stack types include new syntax and new expressions for dealing with
+references on stacks.
+
+## Motivation
+
+Stacks are used in most programming languages in order to pass arguments into
+functions.  Haxe supports this usage natively, with additional extern support
+for features such as Rest, Either, etc.
+
+However, some languages use stacks as return values, and in iteration contexts.
+
+e.g.
+```lua
+ for k,v in tbl : print(k,v) end
+```
+
+If we reference stack-based tuples with a conventional type in Haxe it causes
+all kinds of problems.  @nadako and I worked through some of these for lua
+multireturns, but it's not a first class solution.  Multireturn types cannot
+function as type parameters, and there's currently no Haxe syntax capable of
+generating them (extern only feature).
+
+This problem also crops up with iteration, since many languages support
+iteration over multiple values (python, etc.)
+
+
+## Detailed design
+
+The proposal is to introduce a new expression type representing a stack.  This
+would commonly be used in certain iteration contexts, such as key/value:
+
+```haxe
+for ((k,v) in map){
+  trace('$k: $v');
+}
+```
+
+It would also be possible to generate these stack tuples as return types.
+Languages like Lua could support these natively.  Other languages could support
+these with the help of dummy container values (that are reusable within
+threads).
+
+```haxe
+ (var x,var y,var z) = tupleGenerator();
+// or maybe
+ var (x,y,z) = tupleGenerator();
+```
+
+Stack-based tuples are not just for key/value iteration on certain targets.
+Function argument expressions are also effectively the same thing.  This tuple
+pattern would work well with coroutines that have more than one argument for
+their callbacks:
+
+```haxe
+var some_async_function(cb : Bool->Int->Void){
+
+};
+var (success : Bool ,value : Int) = some_async_function().async;
+```
+
+The paren-based wrapper syntax clearly differentiates stack based datastructures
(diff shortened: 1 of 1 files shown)


jdonaldson:
See also : 
1. https://github.com/HaxeFoundation/haxe/issues/4300
2. https://github.com/nadako/haxe-coroutines
3. https://haxe.org/manual/target-lua-multireturns.html

ncannasse:
I am not sure about it. 
Having used tuples in OCaml for a long time, I don't think they fit well in Haxe design.

You don't really want to have 4+ elements tuples because they are messy and then would be a lot better with proper field names ( = haxe structures ). 3 is a bit in the middle. 2 might be useful, 1 is... well....

So if the question is should we have tuples for the possibility of returning two values in a function, my conclusion would be no.

jdonaldson:
I agree, if this proposal is just about handling 2-value multireturns, it's not worth it. 

Let's see if @nadako thinks this would work for async scenarios.  I think stack based tuples may not get used much for  Haxe code that is written from scratch, but if you're writing a lot of async wrappers for pre-existing callback functions, or if you're writing externs for messy multireturns I think it's a good idea.

ncannasse:
stack based multi return is a bad optimization for us I think, because we have very few platforms that can implement it. So this will lead to some users using them massively and then later noticing they are causing a lot of performance problems on other platforms.

jdonaldson:
I think there's ways to reduce the overhead on platforms that can't implement natively, but I defer to your experience on that point.

Is the mechanism for async handling clear?  I haven't had any feedback on that.  The support there will be the same for all platforms.  If necessary I can split the proposal here and/or refactor.

nadako:
I'm not sure about async stuff atm, but I think Haxe would benefit from tuples a lot. We cannot and we should not make any promises about whether they are stack-allocated or not, but if we say they are non-nullable and read-only, they can be implemented in a native and/or optimized way on many targets, which is not the case with anonymous structures/classes.

As for the usefullness, I think they should have "use-site only" field names, similar to new function type with named argument, so, e.g.:

```haxe
// these two types unify (same arity and types)
typedef T = (a:Int, b:String);
typedef T2 = (c:Int, d:String);

var t1:T = (10, "hi")
trace(t1.a, t1.b);

var t2:T2 = t1;
trace(t2.c, t2.d);
```

Unpacking into local vars is a topic that is not strictly related to tuples, and my current opinion is that we should go for a convention-based destructuring, similar to [one implemented in Kotlin](https://kotlinlang.org/docs/reference/multi-declarations.html), basically:

```haxe
class Pair<T1,T2> {
  public var _1:T1;
  public var _2:T2;
}

// and then
var (a, b) = somePair;

// translated into:
var a = somePair._1;
var b = somePair._2;
```

This convention would then be simply used by tuples, so any tuple type would have `_1`/`_2`/etc. fields.

ncannasse:
I have designed Haxe to be as simple as possible. Simple to read, simple to learn. This means that only the feature that brings something expressive enough to justify adding them to the syntax/language should be considered for addition. Given the amount of overlap between Tuples and anonymous structures, and the ability to generate these tuples with macro, I'm not in favor of adding them to the official language syntax.

bendmorris:
To clarify the status of this proposal - is this a veto by @ncannasse, or a no vote? If the latter, it sounds like the vote may be currently 2 in favor, 1 opposed?

Justinfront:
Is there a repo I can try them out where the c++ and js implementations provide both speed and expressive advantages, along with predictable special cases, I think that should be a minimum before you start counting votes.  If users like me feel they offer an advantage then your halfway there, I already know Tuples so would be keen to see them as part of the language.  Time to demo the implementation, if the theory was enough Nicolas would have agreed?  So initially myself - I only care about js and c++... very pragmatic, flash would be nice but like many of the other targets not the main USP.
benmorris no disrespect but I have only seen your commits for a short time - so if your not happy with Nicolas vote then rather than talk about votes I think you need to get an implement together so we can try it out, seems it's at that stage?

bendmorris:
>so if your not happy with Nicolas vote...

I'm simply asking for an update on the status of this proposal - please don't read any more intent into that.

Justinfront:
Well I am happy to try out a branch if you have one.

jdonaldson:
FWIW this feature would require a lot of buy-in from the rest of the compiler team.   The lack of a stack-indexed datastructure is a minor nuisance now, but it seems like a really good idea for certain async features.  

Also as a side note, I realize not all platforms would support these datastructures in all cases.  They could be emulated with a simple array in most cases I think.

ncannasse:
We might reconsider this while working on Haxe-in-Haxe (OCaml backend etc) but atm I still don't think it would make a good addition for Haxe "core" language. We will eventually experiment it for the OCaml backend if necessary and see how it plays.

Simn:
I want destructuring, but I don't think I want tuples. I also don't want to overload the `( )` syntax any further.

jdonaldson:
I propose we add new bracket delimiters in that case:

var k = 🍺a,b,c 🍺

markknol:
Fwiw, in typescript they use { } for destructuring

     let {x,y} = point;

PR #44, opened 2018-03-18
Complex type expressions

bendmorris:
Adds a syntax to represent types as expressions: `var a = :Array<a.b.MyClass>;`

Discussed in https://github.com/HaxeFoundation/haxe/issues/6910, PR: https://github.com/HaxeFoundation/haxe/pull/6913

[Rendered version](https://github.com/bendmorris/haxe-evolution/blob/complex-type-expr/proposals/0000-complex-type-expressions.md)

### DIFF
--- a/proposals/0000-complex-type-expressions.md
+++ b/proposals/0000-complex-type-expressions.md
@@ -1,0 +1,116 @@
+# Complex type expressions
+
+* Proposal: [HXP-NNNN](NNNN-complex-type-expressions.md)
+* Author: [Ben Morris](https://github.com/bendmorris)
+
+## Introduction
+
+Currently Haxe can express complex types as expressions or metadata arguments if and only if they parse as another type of expression, such as an identifier or field access. This proposal adds new syntax to express complex types which do not fit this criteria.
+
+## Motivation
+
+There is currently some counter-intuitiveness regarding how types can be used as expressions.
+
+Currently types without parameters can be parsed into either a `EConst(CIdent(...))` or a `EField(...)`, and have a corresponding runtime value:
+
+```haxe
+class Test {
+    static function main() {
+        var a = Test;
+        trace(a);
+        // Test.hx:4: Class<Test>
+
+        var b = sys.net.Socket;
+        trace(b);
+        // Test.hx:8: Class<sys.net.Socket>
+    }
+}
+```
+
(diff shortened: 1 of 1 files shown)


nadako:
I'm in favor of this, since I myself had the need to both store type info in metadata and pass it as an argument in a macro function and it always ends up with something ugly like `(_:SomeType)` with subsequent parsing, so it basically asks to have a syntax for that and `:Type` syntax makes sense to me too, because of the way we already use colon for type hinting/checking.

bendmorris:
And note that again this is possible once I add a typedef, because the only barrier is syntax:

```haxe
typedef MyClassInt = MyClass<Int>;

trace(Std.is(e, MyClassInt));       // true
```

So currently we need superfluous typedefs in order to reference a specific `@:generic` class at runtime.

ncannasse:
I don't think that it justify adding new syntax that is very specific to this case.

- for metas you can already sur `var _ : MyClass<String>`  and I think `macro : MyClass<String>` should maybe work as well. That's a good enough workaround
- generic classes requires typedefs, but you can also implement a `@:generic is` or use macros to solve this problem

In both cases, adding new syntax should only be done when it clearly opens new possibilities that wouldn't be possible otherwise. The more syntax a language has the more complex it is to learn/read. Also, using `:` for this case locks it for other future (and better) uses, so I think this proposal should be rejected.

ncannasse:
I agree, but I don't think that justify adding a specific syntax just for
that.

bendmorris:
Fair enough. If we did adopt something that made use of this (such as the `Const<T>` proposal) I think opinions might shift on whether this workaround is acceptable. i.e. I don't think officially requiring:

```
    @:const(var _:Array<Const<T>>) function concat(a:Const<Array<T>>):Array<T>;
```

in user code and the std library would be reasonable.

Simn:
I'm also not really sold on this one. I acknowledge that it makes a specific case a bit nicer, but that alone is not enough to justify adding a new syntax.

back2dos:
Hmm, maybe @FrancisBourre can comment on whether or not this is useful.

We currently use strings and parse those at compile time, e.g. `@type('String -> Int -> Bool')`. We have that a few hundred times in our code base. And we have `@Dependency(var _:SomeType)` another few hundred times.

ncannasse:
I'm voting for a close until someone propose a better solution.

PR #45, opened 2018-04-24
Inlining functions at call location

YellowAfterlife:
Allow requesting a function to be inlined at call location

[Rendered version](https://github.com/YellowAfterlife/haxe-evolution/blob/master/proposals/0000-inline-calls.md)


Simn:
> inline is not currently used in expression syntax at all

That is not _entirely_ true because `inline function` exists at expression level. However, that does not cause any grammar problems.

I'm very much in favor of this. My biggest concern is the required AST change.

ncannasse:
Can AST be parsed with a metadata ? "inline expr" would be equivalent of "@:inline expr" , and we could have the same of "inline function foo()" actually.

Simn:
We could do that, but I think with a major version change we should do it properly and add EInline, similar to EUntyped.

ncannasse:
We would only allow inlining of `@:inline ECall`, not add "inline" meta to whole expression typing as we do for normal metadata.

Simn:
I agree with that.

PR #46, opened 2018-05-17
Intersection types

Simn:
Support `Type1 & Type2` intersection type syntax.

[Rendered version](https://github.com/Simn/haxe-evolution/blob/intersection-types/proposals/0000-intersection-types.md)

### DIFF
--- a/proposals/0000-intersection-types.md
+++ b/proposals/0000-intersection-types.md
@@ -1,0 +1,131 @@
+# Intersection Types
+
+* Proposal: [HXP-NNNN](0000-intersection-types.md)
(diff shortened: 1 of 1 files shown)


sledorze:
I can't emphasis enough to take a look at Typescript definitions of unions / intersections / etc..
Merely because I everybody will want to derive Haxe externs automatically from those Typescript definitions to cover all Web APIs.

ncannasse:
ok for me

nadako:
I find intersection types and unions useful in some situations, and while of course this proposal's functionality is a bit limited, it doesn't seem to restrict future enhancements, so let's do it! \o/

PR #47, opened 2018-05-24
Create 0004-es6.md

jgranick:
ES6 is obviously the future of JavaScript, and is supported by all major browsers. Modern client-side development is designed for modular ES6 output, and macro-based solutions are clumsy.

Let's take our JavaScript to the next generation

### DIFF
--- a/proposals/0004-es6.md
+++ b/proposals/0004-es6.md
@@ -1,0 +1,123 @@
+# ES6
+
+* Proposal: [HXP-0004](0004-es6.md)
+* Author: [Joshua Granick](https://github.com/openfl)
+
+## Introduction
+
+ES6 support for JavaScript in the compiler
+
+## Motivation
+
+JavaScript is getting newer, and better.
+
+Generating classes for ES6 would allow for highly improved integration with
+ES6 source, TypeScript and most modern JavaScript workflows. Also, all major
+browsers support ES6 sources out-of-the-box. This improves debugging, and
+ultimately improves the quality of Haxe's JavaScript.
+
+It's time to move the next generation. Modular JavaScript would benefit highly
+from ES6 JavaScript output. This allows for intelligent bundling, and integration
+with a wide-range of enterprise JavaScript tooling.
+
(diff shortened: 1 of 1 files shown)


EricBishton:
While this proposal adds a new 'define' (or compiler conditional variable definition) to the compiler, it changes the semantics of the current '-js _&lt;filename&gt;_' parameter.  This will cause some difficulties for IDEs and other tools that currently manage the compiler parameters and expect output given the current semantics.  Additionally, a define can be created in many places within the system, beyond the command line, and tools may not have access to them -- or will have to be significantly changed to have such access.  (For example, a define can be created in '--macro' arguments, in actual macros in the source code, in project files, or via the haxelib.json and extraparams.json in libraries.  Since appearances in code can also occur inside of compiler conditional blocks [behind another define] that places a further barrier to knowing whether the define is enabled.)  Not only that, but this would place the meaning of the output flag within the realm of libraries, under which, many, if not most, users prefer not to modify.

Instead, since the output semantics of es6 generation change, I would propose a new target ('--es _&lt;dirname&gt;_' or '--jsmodules _&lt;dirname&gt;_'?), rather than a conditional variable.

To be clear:  This is not a comment upon the desirability or relevance of this proposal -- only the proposed mechanism of achieving it.

demurgos:
I agree that the proposal is desirable. The output code would be more idiomatic and play well with various tools.

As a minor note, it may be good to provide `js-es2015` as an alias: the versioning for versions after ES5 seems to prefer years rather than editions of the spec.

jgranick:
@Simn The most pertinent feature of ES6 for us would be ES6 classes, ES6 import and (in a perfect world) different JS files per module/class.

How would you like me to propose these? Separate issues, or a more specific evolution proposal?

Simn:
See these Haxe issues:

https://github.com/HaxeFoundation/haxe/issues/4552
https://github.com/HaxeFoundation/haxe/issues/5831
https://github.com/HaxeFoundation/haxe/issues/6546

PR #49, opened 2018-07-30
Backticks

markknol:
Support backticks <code>&#96;</code> in Haxe, which also support String Interpolation.

[Rendered version](https://github.com/HaxeFoundation/haxe-evolution/blob/backticks/proposals/0000-backticks.md)

### DIFF
--- a/proposals/0000-backticks.md
+++ b/proposals/0000-backticks.md
@@ -1,0 +1,73 @@
+# Backtick support
+
+* Proposal: [HXP-0000](0000-backticks.md)
+* Author: [Mark Knol](https://github.com/markknol)
+
+## Introduction
+
+Support backticks <code>&#96;</code> in Haxe, which also support String Interpolation.
+
+## Motivation
+
+Escaping quotes in strings can be confusing, but it can be less if Haxe would support backticks.
+
+```haxe
+// best case at the moment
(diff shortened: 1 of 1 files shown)


Simn:
Given that we'll have this inline XML syntax, I don't think this is necessary.

markknol:
I think it still makes sense to have, for safe escaping, localisation and probably for some template engines that don't rely on xml related syntax. 

Question about the inline markup syntax as proposed: Will the syntax (without macros doing stuff with it) compile to strings? The proposal says it is going to trow a `Unprocessed inline markup`, is that true? Or will it sometimes return actual Xml? 

In any case, sometimes it's just easier to write:
```
myElement.innerHTML = `<button onclick="alert('Hi!')">Press me</button>`;
```
.. compared to this:
(below wild guess how the alternative would look with the inline markup syntax (??) ). 
```
myElement.innerHTML = new Xml(<button onclick="alert('Hi!')">Press me</button>).toString();
```
Also since HTML5 isn't valid XML, if you want to write something like `<video autoplay></video>` xml isnt even usable. Or will we have a special processor for html?

kevinresol:
@markknol see https://github.com/HaxeFoundation/haxe/issues/7035#issuecomment-418722413

markknol:
Too bad, thanks for looking into it. Hope the not-XML will help here indeed!

PR #50, opened 2018-08-02
Default type parameters

benmerckx:
Follow up of haxefoundation/haxe#7304

Optionally declare a default type for generic type parameters.

````Haxe
class Test<T = String> {}
$type((null: Test)); // Test<String>
````

[Rendered version](https://github.com/benmerckx/haxe-evolution/blob/master/proposals/0000-default-type-parameters.md)

### DIFF
--- a/proposals/0000-default-type-parameters.md
+++ b/proposals/0000-default-type-parameters.md
@@ -14,6 +14,8 @@ $type((null: Test)); // Test<String>
 
 ## Motivation
 
+### Use case: types
+
 Some types can benefit from having default type parameters. As an example look at
 a generic virtual dom component: 
 ````haxe
@@ -72,6 +74,19 @@ class Image extends Component<js.html.ImageElement>
 But in most other cases you can use `Component` directly without passing a specific element type.
 
 See also: https://github.com/HaxeFlixel/flixel/issues/1677
+
+### Use case: methods
+
+Methods with a generic type parameter are not always able to infer that type from the parameters (especially if that type is optional).
+
+````haxe
+function createMyClass<T = MyDefault>(?input: T): MyClass<T>
+  return new MyClass(if (input == null) new MyDefault() else input);
+
+$type(createMyClass()); // MyClass<MyDefault>
+````
+
+Outlined in more detail [here](https://github.com/HaxeFoundation/haxe-evolution/pull/50#issuecomment-413976704)
 
 ## Detailed design
 
--- a/proposals/0000-default-type-parameters.md
+++ b/proposals/0000-default-type-parameters.md
@@ -15,10 +15,15 @@ $type((null: Test)); // Test<String>
 ## Motivation
 
 Some types can benefit from having default type parameters. As an example look at
-a generic virtual dom component: `class Component<Props: {}, State: {}>` 
+a generic virtual dom component: 
+````haxe
+class Component<Props: {}, State: {}> {}
+````
 Users of the library subclass the Component. Often there's no state or props used,
 but with current haxe it's required to write out the empty parameters explicitly:
-`class MyComponent extends Component<{}, {}>`
+````haxe
+class MyComponent extends Component<{}, {}> {}
+````
 With defaults it's possible to simply extend and leave the defaults to the library:
 ````haxe
 class Component<Props: {} = {}, State: {} = {}> {}
@@ -49,8 +54,8 @@ Don't see any at this time.
 ## Alternatives
 
 - It's possible to emulate with `@:genericBuild` but there's some downsides:
- - Can't use those in macros
- - Can't use the type as a type hint (`var a: MyGenericBuild;`)
+  - Can't use those in macros
+  - Can't use the type as a type hint (`var a: MyGenericBuild;`)
 
 - Parameters can be inferred on first usage, but only when constructing a type.
 
--- a/proposals/0000-default-type-parameters.md
+++ b/proposals/0000-default-type-parameters.md
@@ -1,0 +1,64 @@
+# Default type parameters
+
+* Proposal: [HXP-NNNN](NNNN-filename.md)
+* Author: [Ben Merckx](https://github.com/benmerckx)
+
+## Introduction
+
+Optionally declare a default type for generic type parameters.
+
+````Haxe
+class Test<T = String> {}
+$type((null: Test)); // Test<String>
+````
+
+## Motivation
+
+Some types can benefit from having default type parameters. As an example look at
+a generic virtual dom component: `class Component<Props: {}, State: {}>` 
+Users of the library subclass the Component. Often there's no state or props used,
+but with current haxe it's required to write out the empty parameters explicitly:
+`class MyComponent extends Component<{}, {}>`
+With defaults it's possible to simply extend and leave the defaults to the library:
+````haxe
+class Component<Props: {} = {}, State: {} = {}> {}
+class MyComponent extends Component {}
+class MyComponentWithProps extends Component<MyProps> {}
+class MyComponentWithPropsAndState extends Component<MyProps, MyState> {}
+````
+
+See also: https://github.com/HaxeFlixel/flixel/issues/1677
+
+## Detailed design
+
+- Parse the new type parameter syntax for type declarations
+- Ensure the default unifies with possible type guards
+- Disallow a type parameter with a default to be followed by one without
+- Use the default parameter when the type is used and there's none declared
+
+## Impact on existing code
+
+If the defaults are available in macro context this can break existing macros 
+that work with type parameters. Otherwise code that does not use the defaults
+should function exactly the same.
+
+## Drawbacks
+
+Don't see any at this time.
+
+## Alternatives
+
+- It's possible to emulate with `@:genericBuild` but there's some downsides:
+ - Can't use those in macros
+ - Can't use the type as a type hint (`var a: MyGenericBuild;`)
+
+- Parameters can be inferred on first usage, but only when constructing a type.
+
+- Aliases can be used to set defaults: `typedef EmptyComponent = Component<{}, {}>`
+It usually makes things more complex than necessary, see also: 
+https://github.com/massiveinteractive/haxe-react/blob/19156680859ac0e27249762101cb8533b911a141/src/lib/react/ReactComponent.hx#L14.
+
+## Unresolved questions
+
+- Should the defaults have access to the other parameters?
+  `Test<A, B, C = A & B>` This would introduce a lot of edge cases.
--- a/proposals/0000-default-type-parameters.md
+++ b/proposals/0000-default-type-parameters.md
@@ -32,6 +32,45 @@ class MyComponentWithProps extends Component<MyProps> {}
 class MyComponentWithPropsAndState extends Component<MyProps, MyState> {}
 ````
 
+Another use case is simplifying user facing APIs where some types are only necessary 
+to be given explicitly in very specific cases. The types are ready to be used without
+the user having all of the implementation details.
+
+For example tink_core defines an Error as
+````haxe
+typedef Error = TypedError<Dynamic>;
+````
+where with default type parameters there would not necessarily have been a distinction
+as `Error` and `TypedError` could have been defined as:
+````haxe
+class Error<T = Dynamic> {}
+````
+
+Another example would be a generic Promise implementation which holds error data as well. 
+In most cases it makes a lot of sense to default the error type parameter to an error type
+that works for the user but would not restrict them from using it differently if the use
+case came up.
+````haxe
+class Promise<T, E = Error> {}
+````
+
+Using the dom in the javascript target can also demonstrate the use as accessing elements
+is usually done through `js.html.Element`. That works as long as you want access to those
+properties. But in a few cases you need access to specific properties of the element and
+thus want it typed. Say in a lifecycle method of typical virtual dom components
+(ignoring state or props here):
+
+````haxe
+class Component<E = js.html.Element> {
+  onmount(element: E) {}
+}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by RealyUniqueName** [comment]
That introduces ambiguity for expressions like this:
```haxe
var t = new Test();
```
Currently, it's being typed as `Type<Unknown>` which is later get inferred to a specific type parameter.

benmerckx:
I'm not proposing to change the old behaviour for types without defaults. But yes, if a default was declared it should be used instead of `Unknown` there.

DavisDevelopment:
Oh My Goodness, the number of times I've wished for this! Moreso in methods that accept type parameters than generic types, but both would be super nice. I REALLY need to devote some time to familiarizing myself with OCaml so I can fiddle with the compiler..

benmerckx:
@DavisDevelopment could you share an example of such method? I've not run into enough situations there myself, type inference often suffices, so it's not mentioned in this proposal. It would probably make sense to include it though.

DavisDevelopment:
@benmerckx Yeah, if I'm understanding correctly, then just this should demonstrate:
```haxe
static function magic<In, Out>(input:In, ?convert:In->Out):Out {
  if (convert == null)
    return magic(input, x->x);
  else
    // actually perform the magic
}
```

Simn:
I'm a bit wary of implicit types like that. It can cause some confusion because it's not easy to tell where a type came from. Still, I'm not against this feature from a design point of view.

RealyUniqueName:
@benmerckx 
The example from the proposal:
```haxe
class Component<Props: {} = {}, State: {} = {}> {}
class MyComponent extends Component {}
class MyComponentWithProps extends Component<MyProps> {}
class MyComponentWithPropsAndState extends Component<MyProps, MyState> {}
```
How would I declare `Component<MyState>`? 
If it's not possible, then it's just a partial solution to the problem, which is described in the motivation part.

Also, could you please add some examples for use cases, where you'd want this feature for the reasons other than completely ignoring a type parameter?

back2dos:
> How would I declare `Component<MyState>`?
> If it's not possible, then it's just a partial solution to the problem, which is described in the motivation part.

It's an absolutely adequate solution, that can be expanded upon. Or not, which would be fine.

In languages that don't allow argument skipping for functions, people do the following (as most do in Haxe anyway, because our implicit argument skipping causes quite a lot of headaches): arrange parameters by decreasing likelihood of needing anything other than the default. In this case it's clearly `Props` before `State`, since stateless components with props are far more common than property-less statefull components (which kind of miss the point of composability). See github code search: [391 results](https://github.com/search?l=Haxe&q=ReactComponentOfProps&type=Code) vs [40 results](https://github.com/search?l=Haxe&q=ReactComponentOfState&type=Code).

Now we could use the `default` keyword for explicit skipping, both in function calls and when specifying type parameters, in which case you'd do `Component<default, MyState>`.

> where you'd want this feature for the reasons other than completely ignoring a type parameter?

Uhm, I would start in tink_core with this:

```haxe
enum Outcome<Data, Failure = Error> {
  Success(data:Data);
  Failure(failure:Failure);
}
```

Or simple examples like ```typedef Node<Attributes = haxe.DynamicAccess<String>> = ...`, in which case the default behavior is probably good enough for many cases, but can be specialized to something typed.

I could go on all night, but I don't want to waste your time and mine, so if you could please give a number for how many examples you want, I'd be happy to satisfy your curiosity.

kLabz:
> How would I declare `Component<MyState>`?

Maybe like bind, with `Component<_, MyState>`?

Compared to manually adding the default parameter for `Props` with `Component<{}, MyState>`, this has the benefit of staying up to date if `Component`'s default type parameter changes.

benmerckx:
> How would I declare `Component<MyState>`?
If it's not possible, then it's just a partial solution to the problem, which is described in the motivation part.

In that case you would have a `Component<{}, MyState>`. Skipping default type parameters does not seem like something that is possible without being specific and naming them or using a keyword for ignoring one. That's however not something I'd like to add to this proposal. @back2dos suggested using `Component<default, MyState>` which would work if this is a requirement for having this proposal succeed.

I think the current state of the proposal has enough use without having a mechanism of skipping over some defaults. What I'm proposing is pretty much the same feature as in flow or typescript where it is heavily used.

> Also, could you please add some examples for use cases, where you'd want this feature for the reasons other than completely ignoring a type parameter?

I added a few more. If these are not convincing I'm hoping @back2dos could help me out.

back2dos:
> Maybe kinda like bind, `with Component<_, MyState>`?

Totally forgot about `_` 🤦‍♂️

benmerckx:
Could we have a vote on the proposal as is, excluding skipping of parameters that was mentioned? Skipping parameters doesn't seem like a requirement to me but can definitely be revisited if the need arises.

Simn:
We are fine with the proposal, but have to check how to implement this in our type system. My intuition is that this is similar to constraints, but happens earlier. Not quite sure how much earlier though... do we unify constructor arguments first and then apply defaults, or the other way around?

Simn:
Sorry, I confused proposals. The one I was thinking of was https://github.com/HaxeFoundation/haxe-evolution/pull/36. I don't think we came to a conclusion on this one.

benmerckx:
One more case where this is useful is to help the compiler determine a `@:generic` type which it can't infer:
````haxe
@:generic class Cache<K, V, C = String> {
  final load: K -> V;
  final cacheKeyFn: K -> C;
  public function new(load: K -> V, ?cacheKeyFn: K -> C) {
    this.load = load;
    this.cacheKeyFn = switch cacheKeyFn {
      case null: Std.string;
      case v: v;
    }
  }
  // ...
}
// Currently: could not determine type for parameter C
var cache = new Cache((i: Int) -> i + 1);
````

**Comment by nadako** [comment]
@benmerckx but that looks that it shouldn't compile anyway, because if `C` is anything other than `String`, using `Std.string` is incorrect...

**Comment by benmerckx** [comment]
@nadako If `C` is anything other than `String` it means you are passing a `cacheKeyFn` (not ignoring the optional argument) and therefore not using `Std.string`.

It all boils down to being able to tell the compiler what type a generic type parameter is in case you are manually assigning a default value to an optional (null) argument of which the type cannot be determined. The example might not be super illustrative because it keeps a lot of the details out but it is a use case I ran into.

**Comment by back2dos** [comment]
You could say `new Loader<Int, Int, Int>(i -> i + 1)` in which case the code would still use `Std.string`. The clean solution here would be constructor overloading.

I'm still totally in favor, but this is probably not the best use case ;)

**Comment by benmerckx** [comment]
Right, I also just realized how a cast would be needed there for it to compile with any other `C` (what @nadako said). That makes the example mute.

**Comment by Simn** [comment]

We agreed that this works well for normal type-hints and there shouldn't be any obstacles supporting this feature there.

However, the constructor situation is unclear:

```haxe
class C<T = String> {
	public function new(t:T) {}
}

function main() {
	new C(12); // C<Int> or compiler error?
}
```

The question becomes at which point the default type should be applied.

On a related note, the situation for default type parameters on methods is unclear. In this case, we don't even have a syntax for explicit type parameters. It might make sense to support default type parameters only at type-level, at least initially.

Gama11:
In TypeScript that works:

image

That's the behavior I would expect as well.

back2dos:
I'm leaning toward letting `new C(12)` error. It's a pretty contrived example that diverges from all presented use cases.

The problem that this proposal sets out to solve is that often you have a pattern like this:

```haxe
class FooOf<T:DefaultType> {}
typedef Foo = FooOf<DefaultType>;
```

Where >90% of the time the user will be using `Foo` and in specific cases, they'll want to *narrow* the type - and not substitute it entirely by go from `String` to `Int`. Currently, the cost of this is a 2nd type that needs be remembered and clutters completion.

Consider [this](https://try.haxe.org/#6fF5F):

```haxe
import js.html.*;

class Test {
  static function main() {}
}

class Component<T> {
  public function new(t:T) {}
}

interface Factory {
  function getComponent():Component<Element>;
}

class MyFactory implements Factory {
  var someImage:ImageElement;
  public function getComponent()
    return new Component(someImage);
}
```

This produces an impressive error:

> Test.hx:17: lines 17-18 : Field getComponent has different type than in Factory
> Test.hx:12: characters 2-45 : Interface field is defined here
> Test.hx:17: lines 17-18 : Void -> Component<js.html.ImageElement> should be Void -> Component<js.html.Element>
> Test.hx:17: lines 17-18 : Component<js.html.ImageElement> should be Component<js.html.Element>
> Test.hx:17: lines 17-18 : Type parameters are invariant
> Test.hx:17: lines 17-18 : js.html.ImageElement should be js.html.Element

The point of this proposal is that users who don't care to narrow the type don't have to pay for it being narrowable. With `class Component<T = Element> {}` the above code compile without making the user write `new Component<Element>(someImage)`.

haxiomic:
@back2dos I don't think a default type parameter is the right way to solve the error in your example. The error is fundamentally `js.html.ImageElement should be js.html.Element`, while you can help the compiler by writing `new Component<Element>(someImage)`, I think this inference it should do itself as the return type is already specified to be `Component<Element>` in the interface

When you have 
```haxe
class Component<T> {
    public function new(x:T)
}
```

And you do
```haxe
var anything: SomeType
new Component(anything) // -> Component<SomeType>
```

The default type parameter shouldn't come into play because you've specified the type parameter by passing the argument

The default type parameter should only be used when no type information has been supplied

For example
```haxe
class Vector<T = Float> {
    public function new(length: Int);
    public function get(index: Int): T;
}

new Vector() // -> Vector<Float>
new Vector<Int>() // -> Vector<Int>
```


To implement a component that takes any html element as input, but if none is supplied it generates its own might look like this

```haxe
class Component<T: Element = Element> {
    var element: T

    public function new(?el: T) { }
}
```


So on `new C(12);` I feel strongly it should be `C<Int>`

haxiomic:
@back2dos I don't think a default type parameter is the right way to solve the error in your example. The error is a type-parameter invariance issue `js.html.ImageElement should be js.html.Element`, which I think makes more sense to be solved by improving unification to identify cases where variance is not an issue. Maybe in this case the ideal solution would be:
```haxe
class Component<T: Element> {
    public function new(t: T);
}

new Component(someImage); // -> Perhaps internally the compiler could see this as `new Component((someImage: Element))` because of the constraint
```

When you have 
```haxe
class Component<T> {
    public function new(x:T)
}
```

And you do
```haxe
var anything: SomeType
new Component(anything) // -> Component<SomeType>
```

I argue the default type parameter shouldn't come into play because you've specified the type parameter by passing the argument

The default type parameter should only be used when no type information has been supplied

For example
```haxe
class Vector<T = Float> {
    public function new(length: Int);
    public function get(index: Int): T;
}

new Vector() // -> Vector<Float>
new Vector<Int>() // -> Vector<Int>
```


To implement a component that takes any html element as input, but if none is supplied it generates its own might look like this

```haxe
class Component<T: Element = Element> {
    var element: T

    public function new(?el: T) { }
}
```


So on `new C(12);` I feel strongly it should be `C<Int>`; the default parameter should do nothing to constrain the type of T, only pick a type when none is given

back2dos:
Let's go back to Simn's statement:

> The question becomes at which point the default type should be applied.

There are two options, how to treat omitted optional type parameters:

1. the default is applied
2. the parameter is left open (as it's already the case for mandatory params)

You seem to be suggesting the second option, where the ability to infer the type parameter from constructor usage is simply a fortunate edge case. This however is the actual behavior:

```haxe
class Test {
  static function main() {
    var v = new haxe.ds.Vector(100);
    $type(v);// haxe.ds.Vector<Unknown<0>>
    v[0] = 1;
    $type(v);// haxe.ds.Vector<Int>
  }
}
```
Thus the assertion `new Vector() // -> Vector<Float>` is wrong.

I'm equally confused by `var c: Component<Element> = new Component(someImage);`. It's agreed that for type hints `Component` is fully equivalent to `Component<Element>`. And it thus can be omitted. Which leads us to the following:

```haxe
var c:Component = new Component(someImage);
$type(c);//Component<Element> - ideally this would just print Component
var c = new Component(someImage);
$type(c);//Component<ImageElement>
```

I don't find that particularly intuitive, let alone practical.

Furthermore, you're free to verify that constraining `Component.T` to `Element` does not change the error in the above example I have provided ;)

A default type parameter should be used when said default is useful in the vast majority of cases. It is therefore impractical to then make the user explicitly specify it anyway, to avoid being outsmarted by the compiler.

haxiomic:
I think I didn't fully clarify my examples – none will work in current haxe, they're examples for how things could work 

- On `new Vector() // -> Vector<Float>`, I was suggesting what would happen if you gave Vector a default type parameter of Float, which would resolve the Unknown to Float (which is the behavior in TypeScript)

I'm very aware constraining `Component.T` to `Element` doesn't resolve the error, what I was trying to get at is that with improvements to how constraints are handled it could resolve the error, and the error is fundamentally a type-constraint error and should be fixed that way

The example I gave
```haxe
var c: Component<Element> = new Component(someImage);
```
Was meant to be a reduced example of your case in `getComponent()`, where you have the compiler trying to unify `new Component(someImage)` with the return type of `Component<Element>`. I was suggesting that with improved constraint handling in the compiler, `new Component(someImage)` i.e. `: Component<ImageElement` could be made to unify with `: Component<Element>` (which it currently doesn't because of type-parameter variance)

haxiomic:
I think the thing to do is go over a different use case examples and see if one approach is obviously better than the other

So far I can’t find a strong example for one over another but the explicit (`new C(12)` is an error) approach is more strict and simpler to implement. I imagine we could always extend it to handle inference later if the error becomes a clear problem

Simn:

PR #53, opened 2018-10-26
Clarify instructions re rejecting HXPs

uvtc:
Given [some recent online discussion](https://news.ycombinator.com/item?id=18263505), there was evidently some confusion about what the Haxe-Evolution discussion/voting process is specifically for. I hope this change correctly clarifies that, though please let me know if I've got it wrong here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,18 @@ This project is for maintaining formal change proposals for Haxe language.
 
 ## About proposals
 
-Haxe Proposals (HXP) can be submitted by any Haxe team developer or community member, as long as it's complete and well explained.
+Haxe Proposals (HXP) can be submitted by any Haxe team developer or community member, as long as it's complete and well explained (please see the "How to submit an HXP" section below).
 
 Once a HXP has been submitted, it can be discussed and modified. Anybody can also submit a PR against an existing HXP to have it amended.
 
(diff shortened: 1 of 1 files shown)


uvtc:
Without my extra blockquote (in the PR), the instructions imply that all HXPs will be discussed and voted on. The current instructions can also be easily mis-read to imply that if an HXP is voted down, then Nicolas will explain the reasoning in detail for doing so.

I think it's important to remind folks that the core team is still the arbiter of language design changes, though they're open to community suggestions on the finer points of those changes which they're likely to move forward in making.

Simn:
I don't know, that somehow makes it sound more arbitrary than it is. We actually internally discuss (if briefly) all proposals before rejecting them. It's just that we don't always hold a formal vote because in these cases the result is obvious and we want to be pragmatic about it.

But I agree we should clarify this. I'm just not sure if we can do it in a way that makes everyone happy in case of rejection.

Also, to be perfectly honest, not _all_ proposals are appreciated. Some of them come across as low-effort with little more content than "please add this because XYZ has it too". The whole idea here is that people think things through so we don't have to think so much, because that way we can focus on our strengths instead.

uvtc:
Ok. Updated using your feedback.

I think not everyone is going to be happy with their HXP being rejected and there's any way to avoid that. If I spent a couple of days writing up an HXP and it was flatly rejected, I'd be ok with, "Thanks for your efforts here, but the team discussed this HXP and it's generally not a good fit for Haxe because {insert general issue with the HXP}.". I think it's important to provide some feedback where an HXP goes astray from the guiding principles / goals / vision of Haxe.

uvtc:
What could help here would be a vision document spelling out the core team's vision for Haxe. So, if your HXP aligns with the core team's vision for Haxe, then it's more likely to be discussed, voted on, and accepted.

I asked about this on the forum https://community.haxe.org/t/what-are-the-haxe-language-core-values-philosophy-goals/893 .

PR #55, opened 2019-01-21
Command to init empty Haxe project.

dmitryhryppa:
Providinga a Haxe compiler command to create an empty project in the selected folder.
`haxe --init cpp js hl`

Which will generate something like this:
```
├───.vscode
│   ├── settings.json
│   └── tasks.json
│
├── out
│   ├── cpp
│   ├── hl
│   └── js
│
├── src
│   └── Main.hx
│
├── build.cpp.hxml
├── build.hl.hxml
├── build.js.hxml
├── common.hxml
```

[Rendered version](https://github.com/dmitryhryppa/haxe-evolution/blob/init-command/proposals/0000-init-command.md)

### DIFF
--- a/proposals/0000-init-command.md
+++ b/proposals/0000-init-command.md
@@ -1,0 +1,67 @@
+# Command for empty project initialization
+
+* Proposal: [HXP-NNNN](NNNN-init-command.md)
+* Author: [Haxe Developer](https://github.com/dmitryhryppa)
+
+## Introduction
+
+Providinga a Haxe compiler command to create an empty project in the selected folder.
+
+## Motivation
+
+It is really annoying to create folders structure, files, declare a class and hxml params to just start coding. 
+Also, you need to spend some time to setup VSCode and VSHaxe debuggers (which is an official editor, yes?).
+
+This feature also will help a lot for new developers to start.
+
+## Detailed design
+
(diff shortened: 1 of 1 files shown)


ncannasse:
This should definitely not be part of the compiler, could be added to
vshaxe commands instead.

Le lun. 21 janv. 2019 19:12, Dmitry Hryppa <notifications@github.com> a
écrit :

markknol:
I once created a haxelib called [hxnew](https://github.com/markknol/hxnew) which does something like that, maybe I should add vscode support for it.

Simn:
This is outside the scope of a compiler.

dmitryhryppa:
> This should definitely not be part of the compiler
> This is outside the scope of a compiler.

I thought about TypeScript compiler which has that option, that's why I did this proposal here :) 


> @markknol 
>I once created a haxelib called hxnew which does something like that, maybe I should add vscode support for it.
Or indeed, could be added to vshaxe commands.

Oh, this is a very cool thing. Would be awesome if this can be a part of at least VSHaxe.

fullofcaffeine:
> Oh, this is a very cool thing. Would be awesome if this can be a part of at least VSHaxe.

Or at least mentioned in the haxe.org website. This can help reduce some friction when first trying Haxe.

PR #58, opened 2019-05-06
Promises

Aurel300:
Stepping-stone for reworking the system APIs.

[Rendered version](https://github.com/Aurel300/haxe-evolution/blob/promises/proposals/0000-promises.md)

### DIFF
--- a/proposals/0000-promises.md
+++ b/proposals/0000-promises.md
@@ -1,0 +1,126 @@
+# Promises
+
+* Proposal: [HXP-NNNN](0000-promises.md)
+* Author: [Aurel Bílý](https://github.com/Aurel300)
+
+## Introduction
+
+Add a `Promise` type to the Haxe standard library based on the A+ specification for the most part.
+
+## Motivation
+
+Asynchronous APIs are common in a wide variety of applications. Promises are well-suited for expressing the data flow in such APIs with a small number of idioms (`then`, `all`, etc). Several implementations of Promises exist for Haxe already<sup>[1](https://lib.haxe.org/p/thx.promise) [2](https://github.com/jdonaldson/promhx) [3](https://github.com/haxetink/tink_core)</sup>, as well as the native JavaScript Promise extern `js.Promise`.
+
+Without relying on third-party Promise libraries, asynchronous task completion in Haxe can be facilitated with:
+
+ - `dynamic` callback function (e.g. [`haxe.http.HttpBase.onData`](https://api.haxe.org/v/development/haxe/http/HttpBase.html#onData)) or callback function provided in an argument
+ - synchronous call (the majority of methods in `sys` packages) executed in a separate thread, with some manual state management
+
+Neither of these are particularly clean solutions. Adding a `Promise` type to the standard library would enable providing more unified APIs for all Haxe targets. `sys` classes in particular are good candidates for asynchrony, for example:
+
+```haxe
+// current API
+var done = 0;
+var largeData = ...;
+for (i in 0...3) sys.thread.Thread.create(() -> {
+    sys.io.File.saveBytes('out${i}.bin', largeData);
+    done++;
+  });
+while (done < 3) {
+  // perform other tasks ...
+  Sys.sleep(.25);
+}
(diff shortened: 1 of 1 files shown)


nanjizal:
I think    
```haxe.async.Promise```
Might be better.
Then any tools related to **async** programming could be added at a later date, I believe swift and nodejs are quite async focused so it's feasible that Haxe would add some platform abstraction to provide more consistant journey in future revisions.

**Comment by nanjizal** [comment]
My reasoning is that a Promise is like data structures - ie: non essential more advanced feature, so even ```haxe``` package is a bit high.  For instance if some form of Future was added or async loading approach or async variables then a package signifiies haxe's future intent to support some forms of async development flow.

**Comment by Aurel300** [comment]
> I believe swift and nodejs are quite async focused so it's feasible that Haxe would add some platform abstraction to provide more consistant journey in future revisions.

I am working on that at this very moment, since that is the topic of my Haxe internship :)

As for the `haxe.async` package, it may be a good idea, although I'd like to see what other people think as well.

**Comment by nanjizal** [comment]
Yer yer just adding the idea into the mix.

**Comment by RealyUniqueName** [comment]
I don't like to follow JS here. We had plenty of problems with JS promises. Maybe we need to try and design something specific for the rich Haxe type system.

E.g. what about `Promise<Void>`? 
```haxe
newPromise<Void>((resolve,reject) -> {
    //how to pass Void value to resolve?
    resolve(); //Error: Not enough arguments, expected value:Void
})
    .then(() -> trace('wat'));
```

markknol:
Please consider Promise and Future of tink_core. No need to reinvent the wheel. Also tink_core is really good, well thought.

Aurel300:
> No need to reinvent the wheel.

Isn't that what tink_core did though? :)

But, in all seriousness – the reasons I chose to propose an A+ promise-like interface for the Haxe standard library:

 - Familiarity - A+ promises are the most popular promise interface known, thanks to Node and JS world in general
 - Minimal interface – we just need the one class (`haxe.(async.)Promise`) to be able to provide asynchronous APIs

From what I can see tink_core promises are built on top of tink_core futures and tink_core "surprises", etc etc. However well thought-out these classes are, they are not designed to be taken individually.

@RealyUniqueName I agree there are some problems with typing `Void` promises. Ideally, we would have void values, i.e. `function foo(bar:Void) trace("ok");` could be called with some sort of `foo(voidValue)`. This is most definitely the subject for a separate proposal though.

tink_core promises solve this issue with a workaround of a `Noise` type (`enum Noise { Noise; }`). Even if we went for tink_core-style promises, we would not solve this problem; so that is not an argument in favour of tink_core, rather it is an argument in favour of void-values and/or having a `Noise` type in the standard library.

kevinresol:
My 2 cents about the async api in sys: **Please use callbacks**.

Reasons:
- Familiarity: Callbacks is the most popular and primitive async interface known. 
- Minimal interface: we need no additional types at all, functions are first-class citizens in Haxe
- Least opinionated
- Least overhead (at least architecturally, might not be true for runtime)

nadako:
I also think simple callbacks is a good option for standard library. They can be easily mapped to whatever async operation abstraction the user prefers (which can also depend on the target and interop requirements).

That said I think we should put some thought into whether these callbacks can already be used as "continuations" for our current coroutine design. It would be great if they could, without any further wrapping!

nadako:
Was there already discussion about operation cancellation? I find the ["cancellation token"](https://github.com/vshaxe/json-rpc/blob/master/src/jsonrpc/CancellationToken.hx) approach quite nice for this.

Aurel300:
Well then, after further thought and discussions here and on Slack, I do agree that callbacks are the way to go for the APIs in the standard library. A promisified version of the API may be added as part of a promise implementation later, since it is a question of wrapping the callbacks into promises.

Target-consistent promises would still be a useful addition to the standard library. Maybe *almost* A+ is not the way to go if we actually want to get rid of some of the [A+ problems](https://staltz.com/promises-are-not-neutral-enough.html). I'll close this for now.

PR #61, opened 2019-06-05
Readable/Writbale constraints

lublak:
[Rendered version](https://github.com/lublak/haxe-evolution/blob/readable-writable-constraints/proposals/0000-readable-writable-constraints.md)

### DIFF
--- a/proposals/0000-readable-writable-constraints.md
+++ b/proposals/0000-readable-writable-constraints.md
@@ -1,0 +1,105 @@
+# Readable/Writbale constraints
+
+* Proposal: [HXP-NNNN](NNNN-filename.md)
+* Author: [lublak](https://github.com/lublak)
+
+## Introduction
+
+Combines all readable/writable access types for a constraint check. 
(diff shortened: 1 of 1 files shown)


Simn:
This proposal basically throws a piece of code at us and asks us to make it work. You would have to go more into detail here and describe what you actually want. Is this just syntactic sugar for properties? Or are we talking about variance annotations? Or what?

Closing for now. I'm willing to reopen if clarifications are made.

lublak:
@Simn it was about the idea of ​​readable and writtable constrain checks.
I'm not a native speaker, but I hope you understand what I'm trying to explain.
Currently we have different readable getters (get, never), (default, null) and (default, default).
For a constraints it should not matter how it can be read. Currently if you use <T:{var field:String}> it is impossible to use fields with (get, never), (default, null).
If you use `<T:{var field(default, null):String}>` you only can use `field(default, null)`.
The same applies to writable constrains.

PR #62, opened 2019-06-19
Self access for abstracts

markknol:
Add a way to access "self" for abstracts, which is a getter for `(cast this:MyAbstract)`.

```
abstract MyAbstract(MyType) as <ident> { 
}
```

[Rendered version](https://github.com/markknol/haxe-evolution/blob/abstract-self/proposals/0000-abstract-self-ident.md)

### DIFF
--- a/proposals/0000-abstract-self-ident.md
+++ b/proposals/0000-abstract-self-ident.md
@@ -1,13 +1,13 @@
 # Self access for abstracts
 
-* Proposal: [HXP-NNNN](NNNN-abstract-self-indent.md)
+* Proposal: [HXP-NNNN](NNNN-abstract-self-ident.md)
 * Author: [Mark Knol](https://github.com/markknol)
 
 ## Introduction
 
 Add a way to access "self" for abstracts, which is a getter for `(cast this:MyAbstract)`.
 
-Based on [this discussion](https://github.com/HaxeFoundation/haxe/issues/8162) I'd like to propose an optional `as <indent>` for abstracts, which allows to add/name your own "self":
+Based on [this discussion](https://github.com/HaxeFoundation/haxe/issues/8162) I'd like to propose an optional `as <ident>` for abstracts, which allows to add/name your own "self":
 
 ```haxe
 abstract MyAbstract(MyType) as selfIdent {
@@ -78,12 +78,12 @@ enum abstract Item(String) as item {
 
 ## Detailed design
 
-> `abstract MyAbstract(MyType) as <indent> { }`
+> `abstract MyAbstract(MyType) as <ident> { }`
 
-This feature is only for abstracts. You can name `<indent>` however you want, with exception of existing keywords and it should take field name validation in account. 
-Also, `as <indent>` is completely optional for abstracts.
+This feature is only for abstracts. You can name `<ident>` however you want, with exception of existing keywords and it should take field name validation in account. 
+Also, `as <ident>` is completely optional for abstracts.
 
-If there is already a field with same name as the indent in the scope there should be a duplication error.
+If there is already a field with same name as the ident in the scope there should be a duplication error.
 
  ```
 abstract MyAbstract(MyType) as foo { // line 1: Duplicate abstract field declaration : MyAbstract.foo
@@ -100,7 +100,7 @@ Since it is a new feature and is optional, it doesn't break existing code.
 
 ## Drawbacks
 
-This proposal allows you to name your own indent which is nice and flexible (it allows self being to forwarded from another abstract?) but it doesn't encourage a standard way of writing.
+This proposal allows you to name your own ident which is nice and flexible (it allows self being to forwarded from another abstract?) but it doesn't encourage a standard way of writing.
 This downside is maybe minor, but is one to consider.
 
 ## Alternatives
--- a/proposals/0000-abstract-self-ident.md
+++ b/proposals/0000-abstract-self-ident.md
@@ -85,11 +85,11 @@ Also, `as <ident>` is completely optional for abstracts.
 
 If there is already a field with same name as the ident in the scope there should be a duplication error.
 
- ```
+```haxe
 abstract MyAbstract(MyType) as foo { // line 1: Duplicate abstract field declaration : MyAbstract.foo
   public inline function foo():Void; 
 }
- ```
+```
 
 At the moment, if you forward a field/function and create a same named field,
 there is no duplication/override error. I think we don't have to add error messages for that.
--- a/proposals/0000-abstract-self-indent.md
+++ b/proposals/0000-abstract-self-indent.md
@@ -7,7 +7,7 @@
 
 Add a way to access "self" for abstracts, which is a getter for `(cast this:MyAbstract)`.
 
-Based on [this discussion](https://github.com/HaxeFoundation/haxe/issues/8162) I'd like to propose an optional `as selfIdent` for abstracts, which allows to add/name your own "self":
+Based on [this discussion](https://github.com/HaxeFoundation/haxe/issues/8162) I'd like to propose an optional `as <indent>` for abstracts, which allows to add/name your own "self":
 
 ```haxe
 abstract MyAbstract(MyType) as selfIdent {
@@ -17,9 +17,12 @@ abstract MyAbstract(MyType) as selfIdent {
 
 ## Motivation
 
-I find myself copy/pasting the following piece of snippet quite a lot, when dealing with abstracts.
+I find myself copy/pasting the following piece of snippet quite a lot, when dealing with abstracts:
+```haxe
+var self(get, never):MyAbstract;
+inline function get_self() return (cast this:MyAbstract);
+```
 It would be nice if there was a way to solve this in the language.
-
 
 #### Use case 1 
 One actual example use-case would be this simplified abstract (based on [hx-vector2d](https://github.com/markknol/hx-vector2d/blob/master/src/geom/Vector2d.hx#L89) library).  
@@ -75,26 +78,21 @@ enum abstract Item(String) as item {
 
 ## Detailed design
 
-You can name selfIdent however you want. And it is also completely optional.
+> `abstract MyAbstract(MyType) as <indent> { }`
 
-```haxe
-abstract MyAbstract(MyType) as selfIdent {
-  
+This feature is only for abstracts. You can name `<indent>` however you want, with exception of existing keywords and it should take field name validation in account. 
+Also, `as <indent>` is completely optional for abstracts.
+
+If there is already a field with same name as the indent in the scope there should be a duplication error.
+
+ ```
+abstract MyAbstract(MyType) as foo { // line 1: Duplicate abstract field declaration : MyAbstract.foo
+  public inline function foo():Void; 
 }
-```
-
-This snippet could generate this:
-
-```haxe
-var selfIdent(get, never):MyAbstract;
-inline function get_selfIdent() return (cast this:MyAbstract);
-```
-
-If there is already a `selfIdent` in the scope there should be a duplication error.
+ ```
 
 At the moment, if you forward a field/function and create a same named field,
-there is no duplication/override error, so I think we don't have to add error messages for that.
-
+there is no duplication/override error. I think we don't have to add error messages for that.
 
 ## Impact on existing code
 
--- a/proposals/0000-abstract-self-indent.md
+++ b/proposals/0000-abstract-self-indent.md
@@ -1,0 +1,119 @@
+# Self access for abstracts
+
+* Proposal: [HXP-NNNN](NNNN-abstract-self-indent.md)
+* Author: [Mark Knol](https://github.com/markknol)
+
+## Introduction
+
+Add a way to access "self" for abstracts, which is a getter for `(cast this:MyAbstract)`.
+
+Based on [this discussion](https://github.com/HaxeFoundation/haxe/issues/8162) I'd like to propose an optional `as selfIdent` for abstracts, which allows to add/name your own "self":
+
+```haxe
+abstract MyAbstract(MyType) as selfIdent {
+  
+}
+```
+
+## Motivation
+
+I find myself copy/pasting the following piece of snippet quite a lot, when dealing with abstracts.
+It would be nice if there was a way to solve this in the language.
+
+
+#### Use case 1 
+One actual example use-case would be this simplified abstract (based on [hx-vector2d](https://github.com/markknol/hx-vector2d/blob/master/src/geom/Vector2d.hx#L89) library).  
+Test on https://try.haxe.org/#5979b (which obviously has handwritten `self`)
+
+```haxe
+private typedef Vector2dImpl = { x:Float, y:Float }
+
+@:forward
+abstract Vector2d(Vector2dImpl) as self from Vector2dImpl {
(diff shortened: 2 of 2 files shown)


nadako:
I like the idea and I can totally relate to the use-cases. One thing I'm not sure about is whether we should actually generate a getter field. I would probably go with adding an identifier resolution rule instead.

markknol:
I'm obviously no compiler developer so it doesn't have to be a actual field. As long as it works that way. 
What's a better way to describe that particular section?

nadako:
> I'm obviously no compiler developer so it doesn't have to be a actual field. As long as it works that way.
> What's a better way to describe that particular section?

I mean, that is also an option, I'm just proposing an alternative that seems a bit more clean to me (however it might be more dirty implementation-wise). If you want to also add this to the proposal, then I'd say something like:

> Alternatively there could be a new identifier resolution rule:
> * When resolving and identifier while typing a non-static abstract method, consider the name, provided with `as <ident>`.

Although this raises more questions, as in at what place in the resolution order should it be considered, can it be overshadowed by local vars, should there be an error when declaring an argument/local var/field with the same name, etc.

RealyUniqueName:
I think it should behave the same way as a field, but generate the same syntax tree as implicit cast of `to Type`.

Simn:
This is a good idea.

I have a question about the syntax though: Shouldn't the `as` be part of the underlying type? For instance, `abstract Vector2d(Vector2dImpl as self)`. The reason is that we're referring to the underlying type, not the abstract. This would also avoid a special case for `@:coreType` abstracts where we have no underlying type.

> I think it should behave the same way as a field, but generate the same syntax tree as implicit cast of from Type.

I agree.

benmerckx:
Isn't `this` referring to the underlying type, whereas `self` or whichever identifier would refer to it typed as the abstract?

Simn:
You're right, I have no idea what I was thinking.

haxiomic:
I like the concept well enough but just a nitpick: I struggle with the `as <ident>` syntax – It took me a while to figure out what this meant. I associate `as X` with X being a type, and it was unexpected that it actually referred to a new member ident because I've never seen this before

Maybe we don't need the ident to be variable – we could have a standard ident that's available in abstract contexts, something like `thisAbstract`

If [onsite getters](https://github.com/HaxeFoundation/haxe-evolution/pull/63) land then the original code snippet would be simpler
`var self(() -> (cast this: MyAbstract), never):Bool;`, but I still see the value in making this a standard feature

RealyUniqueName:
> I associate as X with X being a type

What about `in`? 
`abstract MyAbstr(MyClasss) in <ident>`

RealyUniqueName:
Also let's do the same thing for the underlying type:
```
abstract MyAbstr(MyClass in <identUnderlying>) in <identAbstract> {
```
because similar trick is also used a lot if an abstract doesn't have `to UnderlyingType` but accepts other instances of that abstract as methods arguments:
```
abstract Money(Int in int) {
    @:op(A + B) @:commutative static function add(a:Money, b:Money):Money {
        return a.int + b.int;
    }

    public function new(amount:Int) {
        this = amount;
    }
}
```

Gama11:
I don't see how using `in` helps. It doesn't distuingish it from types because there's already the quasi-deprecated `import <Type> in <Alias>;` syntax. I think in both cases it sounds more awkward than `as`.

kevinresol:
Can we just use the `abstract` keyword?

ncannasse:
I'm okay with the feature but it all depends on the syntax. I would not like another magic identifier so it requires some kind of declaration. Maybe something like

@:this var self(get,never) : MyAbstract;

nanjizal:
I think my use case is related...
I would like to use Array Access internally within an abstract when abstracting an Array, Haxe seems fine with **this[0]** but not with **[0]**, the second I am told refers to self which is not supported?
The problem abstract code ..
``` haxe
public var red( get, set ): Float;
    function get_red(): Float {
        return [ 0 ];
        //return read( 0 );
    }
    function set_red( v: Float ): Float {
        [ 0 ] = v;
        //write( 0, v );
        return v;
    }
```
For example and more context, this experiment works, but I had to call methods because within the abstract it did not seem possible to use array access: 
   http://try-haxe.mrcdk.com/#d6192
Instead it was suggested over the haxe discord channel that I use _self_ by Gama11:
    http://try-haxe.mrcdk.com/#e71a1
_( rem to switch haxe ver in compiler options )._
But if you notice in the generated Javascript code - there seems more overhead with a self getter/setter than in my original code, self is not optimum unless the compiler does some magic to reduce it back to my original read and write methods.

From a users perspective it would seem logical that an abstract should allow me to clarify between **self** and **this**, I really don't like some obscure @ macro stuff. 

I am pretty sure if I understand Nicolas, his proposal will not be something I will forget and I will spend half an hour every time I need to use it trying to work out which subsection of the abstract documentation covers it's use or which pull request references it.  I think it is far more natural to have a magic 'self'.... - it's needed for any advanced abstract use, further it should not ( ideally ) add a function call overhead as a getter setter seems to - lots of compiler magic is ideal here!!

From a language design perspective abstracts are great they allow you to structure your data more like a 'Data Orientated Design' so for example I was exploring storing my WebGL positions and colours in Arrays but then using abstracts to allow easier and cleaner access and creation.

I think 'self' as a keyword should be added since abstracts need them, _self_ seems pretty clear term. Also I would still like to be able to iterate over abstract enums without having to google for a macro :)

anyway just my feelings on self perhaps they will provide different perspective.

Simn:

We agree that this feature is useful, but the syntax remains an open question.

nanjizal:
I think generally abstracts suffer from a lot of 'this' compared to classes could we just use "_" to reduce noise, I think it would be really easy to remember.
```
_[0]
this[0]
```
For any cases where _ is not clear where it may conflict with other _ use the user can just define a local var however they want.
```
var here =_;
var abstract =_;
var self =_;
```
and if it has a parent abstract you can access that via
```
var parentAbstract = _.super;
var grandParentAbstract = _.super.super;
```
Well just adding ideas as really keen on an implemention, so maybe it will sporn other ideas.

nanjizal:
fair enough if your not keen on that suggestion, but I don't want self problem to be ignored.

how about allowing the user to define the keyword, but they have to define it if they need self, this is not too hard to remember and it similar style to forward.
```
@:self('myInt') 
@:forward
abstract MyInt( Int ) from Int to Int {
```
Technically this is useful because you could access 'myInt' from within a child inherited, but would need to error if an inherited abstract tried to reuse the same self.
```
@:self('mySpecial') 
@:forward
abstract MyIntSpecial( MyInt )
```

nadako:
Let's not overload `_` with even more meanings please :)

markknol:
@nanjizal You can always do `abstract MyAbstract(MyType) as _` if you really want to use that 😉 That's the nice thing, you can name it however you want. However if you want `super` in abstract, that would require another proposal I guess. But I personally didn't have a need for that so far, because I don't have such deep layers of abstractiness.

kevinresol:
Just wanna bump the idea of using the `abstract` keyword itself.

`abstract` is now/potentially used as:
- type declaration (`abstract Foo(String)`)
- type modifier (`abstract class`)
- field modifier (`abstract function`)

I think there will be no problem if we use it in expression-level for self-access? This won't break anything because even in macro it would just be `EConst(CIdent('abstract'))`, and it requires no addition syntax to declare (e.g. `as/in/@:this` as proposed by others).

What is the exact reason to reject this syntax proposal?

RealyUniqueName:
We also want something for accessing `this` on an abstract instance. I guess if we go for `abstract` as proposed by Kevin, then we can use `.this`.
I just don't like the keywords to be overloaded with even more different meanings.

nanjizal:
I tried to find some other words that might be short that relate to **abstract**, **this** or similar.
```
virtual
unreal
notional
here
that
hereof
```
I think **hereof** is quite a fitting term, often used by Lawyers, it it easy to google, and explain. 
It is likely users may already use 'here' so here on it's own should be avoided.

The down side on using **abstract** is that it gets more complex to document with it having many meanings, and easy for new users to confuse.

Obviously there are **here..** variants 
http://www.plainenglish.co.uk/the-a-z-of-alternative-words.html

An interesting one would be **hereunder** which could be used consistently to access a parent abstract.

Anyway I am just providing possibles, it's true haxe is keen to use **abstract** term for quite a few uses, it maybe wise not to use it up. 

At least with **hereto** there is no alternate language programming expectation that needs to be corrected.

Aurel300:
How about:

```hx
abstract Foo(anyIdent:Int) {
  ...
}
```

haxiomic:
@Aurel300 that's a pretty neat approach, seems clear enough to me

kLabz:
> How about:
> 
> ```hx
> abstract Foo(anyIdent:Int) {
>   ...
> }
> ```

Seems weird to me to have `this` be of type `Int` and `anyIdent` of type `Foo` though.. (but inverting would be a brutal breaking change..)

emugel:
@Aurel300 I think this suffers the same problem as Simn proposal above. 
The `self` doesn't refer to the underlying type since it is already referenced by `this`.
 
Therefore `abstract Foo(anyIdent:Int) {}` feels wrong to me because it seems to associate `anyIdent` with `Int`, while the discussion is about associating it to `Foo`.

---

Personally I see no big problem with the original proposal, 

`abstract MyAbstract(MyType) as <ident> {}`

it can be misread by a newbie (confusion with `import __ as __`), but a newbie will have to read the doc when he stumbles upon `abstract` anyway. 

But:
* damn easy to remember
* damn fast to use and convenient! 

(I didn't know the trick detailed in the rendered version, underusing the abstract feature because of that, so I'm very much in favor of some simple solution).

emugel:
Alternatively, since abstracts requiring a `self` access are far from being the majority, how about a meta like `@:abstractSelf(self) abstract Foo (String) {}`?

This is self-documenting and awareness of that feature can be maximized by a simple paragraph in the documentation.

kevinresol:
> abstracts requiring a `self` access are far from being the majority

Just to clarify that self access is not uncommon

and wanna raise again the `abstract `keyword, which is not used in expression level right now and will involve no breaking change to macro API (just add `CIdent('abstract')`)

```
abstract Error(Int) {
	var Forbidden = 403;
	var NotFound = 404;
	// ...
	
	public function toString() {
		switch abstract {
			case Forbidden: 'Forbidden';
			case NotFound: 'Not Found';
		}
	}
	
	public function foo(bar:Int) {
		abstract.bar(); // to resolve shadowed field
	}
	
	public function bar(v:Int) {}
}
```

emugel:
Ah suddenly I understand it with your example.

Find the abstract is an elegant solution, but it may be worth posting more examples e.g. with operator overloading because it falls into place so neatly all the question should revolve around the aesthetics with this solution: is it shocking or not.

emugel:
@kevinresol 
After trying some `abstract` types using `abstract` special variable, even though I initially loved the idea, I think there are a few problems, mostly psychological ones:

1. This kind of special variable is usually short (`this`, `that`, `self` are all 4 characters) while `abstract` has the double, 8 characters, bringing uneasiness. Programmer's reflexes won't work as effectively if its instinct is countered.
2. In something like `switch abstract { ... }`, because of the connotations of `abstract` usage in Haxe, it potentially conveys the notion that the following block is some "abstract" thing, not a special variable. Ok this impacts the beginner only but... psychology is also at work here as in the first point.
3. Who's to say that one day, we won't have a use for `abstract { some block }` or `abstract { some anon}`? *special keyword* VS *special variable name*).

So even if `abstract` seems to be free in this context, and it seems a good opportunity to use it, in my mind it is better to not use it this way.

TheDrawingCoder-Gamer:
i really like the `abstract` idea. why not reuse a keyword we already have? And it makes more sense imo. You access this abstract with `abstract`. About your child abstracts, that's tricky. I don't know what you should use. Maybe `this.abstract` ? And for the super super idk. this is just silly at this point. `this.super.super` could work but I'm unsure I like that syntax.

Aurel300:
Let's have a summary of all the ideas so far… For an abstract type `Abs` defined over `T` (`abstract Abs(T)`):

| Declared as | Local access to `Abs` | Local access to `T` | Instance access to `x: Abs` | Instance access to `x: T` | Breaking change | Source |
| --- | --- | --- | --- | --- | --- | --- |
| `abstract Abs(T) as ident` | `ident` | `this` | `x` | - | N | [original proposal](https://github.com/markknol/haxe-evolution/blob/abstract-self/proposals/0000-abstract-self-ident.md) |
| - | `self` | `this` | `x` | - | N | [original proposal, alternatives](https://github.com/markknol/haxe-evolution/blob/abstract-self/proposals/0000-abstract-self-ident.md) |
| - | `thisAbstract` | `this` | `x` | - | N | [comment](https://github.com/HaxeFoundation/haxe-evolution/pull/62#issuecomment-530480731) |
| `abstract Abs(T) in ident` | `ident` | `this` | `x` | - | N | [comment](https://github.com/HaxeFoundation/haxe-evolution/pull/62#issuecomment-530486700) |
| `abstract Abs(T in identT) in identAbs` | `identAbs` | `this` (or `identT`) | `x` | `x.identT` | N | [comment](https://github.com/HaxeFoundation/haxe-evolution/pull/62#issuecomment-530487087) |
| - | `abstract` | `this` | `x` | `x.this` | N | [comment](https://github.com/HaxeFoundation/haxe-evolution/pull/62#issuecomment-530555530), [comment](https://github.com/HaxeFoundation/haxe-evolution/pull/62#issuecomment-647362017) |
| `@:this var ident(get, never): MyAbstract` | `ident` | `this` | `x` | - | N | [comment](https://github.com/HaxeFoundation/haxe-evolution/pull/62#issuecomment-547308811) |
| - | `_` | `this` | `x` | - | Y | [comment](https://github.com/HaxeFoundation/haxe-evolution/pull/62#issuecomment-646866358) |
| `@:self("ident") abstract Abs(T)` | `ident` | `this` | `x` | - | N | [comment](https://github.com/HaxeFoundation/haxe-evolution/pull/62#issuecomment-646894602) |
| - | `virtual`, `unreal`, `notional`, `here`, `that`, `hereof` | `this` | `x` | - | Y | [comment](https://github.com/HaxeFoundation/haxe-evolution/pull/62#issuecomment-647498074) |
| `abstract Abs(identT: T)` | - | `this` (or `identT`) | `x` | `x.identT` | N | [comment](https://github.com/HaxeFoundation/haxe-evolution/pull/62#issuecomment-708391639) |
| `@:abstractSelf(ident) abstract Abs(T)` | `ident` | `this` | `x` | - | N | [comment](https://github.com/HaxeFoundation/haxe-evolution/pull/62#issuecomment-710052129) |
| - | `thisAbs` | `this` | `x` | - | Y | [comment](https://github.com/HaxeFoundation/haxe-evolution/pull/62#issuecomment-950149968) |
| `abstract Abs(identT: T)` | `this` | `identT` | `x` | `x.identT` | Y | [comment](https://github.com/HaxeFoundation/haxe-evolution/pull/62#issuecomment-950311922) |
| - | `this.this` | `this` | `x` | - | N | [comment](https://github.com/HaxeFoundation/haxe-evolution/pull/62#issuecomment-950312478) |
| - | `super` | `this` | `x` | - | N | [comment](https://github.com/HaxeFoundation/haxe-evolution/pull/62#issuecomment-950646197) |

Simn:

PR #63, opened 2019-08-30
 Onsite getters and setters implementation

dmitryhryppa:
Add a way to implement getters and setters directly in the place where they are declared. At least for simple (one line) expressions.

Example:
```haxe
public var isDataExists(() -> data != null, never):Bool;
```

[Rendered version](https://github.com/dmitryhryppa/haxe-evolution/blob/feature/onsite-getset-impl/proposals/0010-onsite-getset-impl.md)

Thanks!

### DIFF
--- a/proposals/0010-onsite-getset-impl.md
+++ b/proposals/0010-onsite-getset-impl.md
@@ -1,0 +1,70 @@
+# Onsite getters and setters implementation
+
+* Proposal: [HXP-NNNN](NNNN-onsite-getset-declaration.md)
+* Author: [Dmitry Hryppa](https://github.com/haxedev)
+
+## Introduction
+
+Add a way to implement getters and setters directly in the place where they are declared. At least for simple (one line) expressions.
+
+Example:
+```haxe
+public var isDataExists(() -> data != null, never):Bool;
+```
+
+## Motivation
+
+Sometimes we need to implement getters or setters with a very simple implementation, just to encapsulate and protect some data.
+
+Nowadays we have short arrow functions which have been added in Haxe 4. So, it would be nice to have the ability to use them with getters and setters to reduce boilerplate code.
+
+## Detailed design
+
+Proposed "short syntax" of getters and setters implementation may work pretty nice with brand new arrow functions. Let's compare the next examples:
+```haxe
+public var isDataExists(get, never):Bool;
+
+//and somewhere in the class:
+function get_isDataExists():Bool {
+	return data != null;
+}
+```
+VS
+```haxe
+public var isDataExists(() -> data != null, never):Bool;
+```
+
+This feature may be done without `() -> expr`, with just an expression inside. But it may be a bit less consistent compared to a full syntax.
+```haxe
+public var isDataExists(data != null, never):Bool;
+```
+
+## Impact on existing code
+
+Does not impact.
+
+## Drawbacks
+
+.
+
+## Alternatives
+
+Alternatively, it may be implemented with a macro.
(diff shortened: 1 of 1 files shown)


RealyUniqueName:
This is could be handy especially for abstracts.

What would be the syntax to override such getters/setters in child classes?

dmitryhryppa:
> What would be the syntax to override such getters/setters in child classes?

Well, It may be parsed as `final` get\set function so it cannot be overridden by child entities.

Gama11:
This looks pretty much exactly like https://github.com/HaxeFoundation/haxe/pull/7129.

>What would be the syntax to override such getters/setters in child classes?

...and bringing up override syntax is one of the things that made @simn reconsider.

dmitryhryppa:
I've added notes about overriding in drawbacks section.

nadako:
I wanted something like this for a long time, but I can't shake of the feeling that what was proposed so far is not very useful in practice...

E.g. in the mentioned case of `abstract` properties, most often it's about fancy field forwarding which you want to be `inline`, but this proposal doesn't seem to support defining `inline` accessors, which makes it less useful.

RealyUniqueName:
Maybe such syntax should imply `inline`, which kind of solves the design for overriding.

RealyUniqueName:
But that doesn't sound very obvious...

dmitryhryppa:
Haxe allows as to use `inline` keyword on static variables. Why not allow it for variables with get/set as well?

```haxe
//will be inlined both(get and set), cannot be overriden:
public inline var isDataExists(() -> data != null, never):Bool;

//Not inlined, can be overriden:
public var isDataExists(() -> data != null, never):Bool;
```

And for full syntax:
```haxe
//will be inlined both(get and set), cannot be overriden:
public inline var isDataExists(get, never):Bool;
function get_isDataExists():Bool { // inline
    return data != null;
}

//Which is the same as:
public var isDataExists(get, never):Bool;
inline function get_isDataExists():Bool {
    return data != null;
}
```

ncannasse:
I don't think we should accept things that are just syntactic sugar and complexify the language while not bringing any new actual feature, especially if the main goal is just to save a function declaration.

nadako:
I tend to agree with Nic. At the same time I wouldn't be opposed to a shorter getter/setter definition, but I think that trying to shove accessor functions betwee name and type won't work. If anything, we could maybe think about "proper" syntax, something along these lines:
```
var prop:Int {
	inline get: _prop, // generates `inline get_prop():Int return _prop`
	override set: _prop = value // generates `override set_prop(value:Int):Int return _prop = value`
};
```

dmitryhryppa:
The short syntax is about how to simplify things and not about how to add more complexity into language. As for me, it's more about options and flexibility. 

But! I don't have power here. Just saying :)

kLabz:
+1 for C#-like [properties](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/properties) which can be used with blocks too (just to keep things in the same place).

As for access control, we can use `private`, `inline`, `override` on `get` / `set`

Simn:
FWIW I'm also in favor of C#-like properties.

Simn:

We mostly agreed that something like this would be good, but it's unclear if it should work exactly as presented here.

Simn:
lean-reject: proposal too weak, feature itself might be valid but needs more thought and design

Simn:

While we agree that a shorthand syntax for properties would be nice, we don't think the proposed one is ideal. There was general agreement among the team that something like the C# property syntax would be better. A proposal which aims to support this would appear to have a high likelihood of being accepted.

PR #64, opened 2019-08-30
Replace "null" access modifier with "this"

dmitryhryppa:
It's a bit confusing to use `null` as a private access modifier.
`this` is much more understandable and gives as a context about who can modify the current field.

Example:
```haxe
public var data(default, this):Bool;
```

[Rendered version](https://github.com/dmitryhryppa/haxe-evolution/blob/feature/access-modifier-change/proposals/0010-null-access-modifier.md)

Thanks!

### DIFF
--- a/proposals/0010-null-access-modifier.md
+++ b/proposals/0010-null-access-modifier.md
@@ -1,0 +1,42 @@
+# "null" access modifier vs "this"
+
+* Proposal: [HXP-NNNN](NNNN-null-access-modifier.md)
+* Author: [Dmitry Hryppa](https://github.com/haxedev)
+
+## Introduction
+
+Replace `null` access identifier with `this`.
+
+## Motivation
+
+It's a bit confusing to use `null` as a "protected" access modifier.
+`this` is much understandable and gives as a context about who can modify the current field.
+
+## Detailed design
+
+It's only about replacing a `null` keyword with `this` as a "protected" access modifier.
+
+Example:
+```haxe
+public var isDataExists(default, this):Bool;
+```
+
+## Impact on existing code
+
+It will conflict with the old codebase. But `null` may be marked as deprecated for some time before it will be fully removed from access modifiers syntax.
+
+## Drawbacks
+
+.
+
+## Alternatives
+
+.
+
+## Opening possibilities
+
+.
+
+## Unresolved questions
+
+.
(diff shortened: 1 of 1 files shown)


ncannasse:
We could go for this, as long as `null` is still accepted and does not generate a deprecation warning. We could replace `null` by `this` in official documentation (and still state that `null` is no longer the spec, although accepted for backward compatibility)

kLabz:
Why `this` over `private` or `protected`?

dmitryhryppa:
> Why this over private or protected?

`private` in Haxe may be confusing because `private` field can be modified by child classes. While `null` access means "truly private" when only the owner can modify that field.  

`protected` in other world is mean the same what `private` in Haxe do, so it is may be confusing too :)

nadako:
> `private` in Haxe may be confusing because `private` field can be modified by child classes. While `null` access means "truly private" when only the owner can modify that field.

I don't think that's true. But @ncannasse had some other argument against `private`. Something about mixing visibility and access modifiers.

dmitryhryppa:
> I don't think that's true.

Oh damn :D

Simn:
I'm against making such small changes to property syntax. It already is a mess and allowing more keywords only makes it even more of a mess. I also don't find `this` more intuitive than `null`, so in both cases it's a matter of remembering the right keyword.

I would leave the current property syntax alone and focus on introducing an entirely different alternative akin to C#.

Simn:
... which I just noticed is also being suggested in #63.

emugel:
It’s not untrue, what Simn says. 

Besides despite me hating `null` as an access modifier, for me properties are not the best part of Haxe and it would be putting new paint rather than thinking it over. 

One grip I have with Properties is they make magic until the don’t (e.g. you can’t override them). 

Once we get the new keyword, it will be less ugly than `null` for sure, but what will you do with the existing codebase? Convert it all? Only use it in future developments? Have a little bit of `this` a little bit of `null`. This is fooling around. It’s better to keep the same old property notation sustem rather than be fooled by some novelty keyword bringing appearance of improvement when in fact the limitations remain.

Simn:
This has been rejected by votes from @ncannasse, @nadako, @RealyUniqueName, @jdonaldson and @simn for the reasons previously stated. 

#63 is unaffected by this decision.

PR #65, opened 2019-09-06
Carry comments through to target language (through AST)

TomByrne:
A more comprehensive representation of code comments in Haxe AST would make for much more readable output and facilitate people using Haxe as a source-to-source tool.

Using a compiler flag, comments could be included in the generated code (e.g. `-Dcomments`) .

This would replace the `doc` property that currently exists on `BaseType`, `ClassField` & `Field`.

[Rendered doc](https://github.com/HaxeFoundation/haxe-evolution/blob/772e4a27a48d56595225aafdbaa94156c675a8f1/proposals/0000-comments-in-ast.md)

### DIFF
--- a/proposals/0000-comments-in-ast.md
+++ b/proposals/0000-comments-in-ast.md
@@ -33,8 +33,8 @@ enum Comment
     Multi(value:String); 
     
     /**
-	* Javadoc style example
-	*/
+    * Javadoc style example
+    */
     Doc(value:String)
 }
 ```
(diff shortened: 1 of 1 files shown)


Simn:
I recommend splitting up this proposal into two parts:

1. Retain documentation comments on types and fields.
2. Retain in-code comments.

The reason is that 1. is several orders of magnitude easier to support. In most generators, it should come down to emitting an additional print at the right places (although determining these places might be a bit finicky).

I think we are unlikely to accept the second part because the benefit/complexity ratio is dubious.

TomByrne:
Sounds good.
If you don't think it'll ever be accepted, I'll just remove the section regarding comments in-code.
Any thoughts on enhancing the parsing of comments?

Simn:
We will implement support for retaining type and field comments in the output, see linked Haxe issue.

I'll close this PR because the actual proposal is a bit wider. It's unclear if in-code comments are feasible, but this can be discussed at a later point.

PR #74, opened 2020-05-26
Local Variable Metadata

Antriel:
Add new syntax for metadata on local variables: `var @:meta foo:Bar;`.

[Rendered Version](https://github.com/Antriel/haxe-evolution/blob/master/proposals/0000-local-var-metadata.md)

### DIFF
--- a/proposals/0000-local-var-metadata.md
+++ b/proposals/0000-local-var-metadata.md
@@ -25,9 +25,11 @@ I propose a new syntax for local variables:
 var @:meta foo:Bar;
 ```
 
+This syntax is similar to already supported method arguments' metadata `function foo(@:meta bar:Bar) {}`.
+
 `haxe.macro.Type.TVar` returned from `haxe.macro.Context.getLocalTVars()` already has `meta` property that could be filled with this new syntax. Currently it's always empty, because normal syntax of `@:meta var foo:Bar;` adds the metadata on the expression, not the variable declaration.
 
-This syntax shall work regardless of whether the variable has an explicit type, and/or initialization, and shall work with the comma syntax. All those shall be therefore valid:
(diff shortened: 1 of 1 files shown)


Simn:
I'm in favor of this, but at least two things still need clarification:

* Please mention that this also works with the comma syntax, i.e. `var @:meta x, @:meta y`
* Please draw a comparison to metadata function arguments, which should offset the "different from class fields" point.

Antriel:
First point added, not sure what exactly you mean by the second one. What are metadata function arguments?
Do you mean an example comparing the currently available solution (moving the metadata up to method level) to how it could look like with the new syntax? I don't see how that offsets the "different from class fields" point.

Simn:
I mean that `function test(@:meta x) {}` is already valid syntax and aligns nicely with this proposal.

Antriel:
Added. I didn't even know about that one (or forgot, regardless, I didn't find it in the manual). :)

haxiomic:
Maybe it would make sense for the alternative syntax to also work (`@:meta var foo: Bar`), because I think that's the one users will likely try first when they add meta to variables

Antriel:
I agree. I like that solution more.
As I outlined, maybe it could be implemented by duplicating the metadata of var declaration expression into the var itself. I am not sure how difficult, if even possible, it is. I'm also not sure what consequences it might have to existing code (I can't think of any though).

Simn:
No, we won't change it like that. It's expression metadata and shouldn't become a special case for local variables. It also doesn't play well with comma syntax.

haxiomic:
Ahh agreed, I misunderstood the proposal initial

Simn:
We have decided to accept this proposal in our haxe-evolution meeting yesterday.

There's not much to say about this one. Progress will be tracked in the linked Haxe issue.

PR #78, opened 2020-07-18
From-to shortcuts

vonagam:
Syntax for allowing from and to cast:

```haxe
abstract Foo(Int) from to Int {}
// equals
abstract Foo(Int) from Int to Int {}
```

Syntax for referencing underlying type:

```haxe
abstract Foo(Int) from this {}
// equals
abstract Foo(Int) from Int {}
```

Obviously for such short named types as std `Int` it doesn't make much sense, but for long named ones and for unnamed, like structures or functions, it would make a life more pleasant.

[Render.](https://github.com/vonagam/haxe-evolution/blob/from-to-shortcuts/proposals/0000-from-to-shortucts.md)

### DIFF
--- a/proposals/0000-from-to-shortucts.md
+++ b/proposals/0000-from-to-shortucts.md
@@ -1,0 +1,47 @@
+# From-to shortcut
+
+* Proposal: [HXP-NNNN](NNNN-from-to-shortcuts.md)
+* Author: [Dmitrii Maganov](https://github.com/vonagam)
+
+## Introduction
+
+New short syntax for allowing both from and to cast.  
+New syntax for referencing underlying type in a cast.
+
+## Motivation
+
+Reduced repetetion and improved readability for two really common scenarios.
+
+## Detailed design
+
+Syntax for allowing from and to cast:
+
+```haxe
+abstract Foo(Int) from to Int {}
+// equals
+abstract Foo(Int) from Int to Int {}
+```
+
+Syntax for referencing underlying type:
+
+```haxe
+abstract Foo(Int) from this {}
+// equals
+abstract Foo(Int) from Int {}
+```
+
+## Impact on existing code
+
+None.
+
+## Drawbacks
+
+Some might say that one can forget a type between `from` and `to`...
+
+## Alternatives
+
+Repetetion.
+
+## Unresolved questions
+
+Which variant to choose for from/to cast.


emugel:
`from to Int` feels oddly like it's missing something...
`abstract Foo(Int) tofrom Int {}` would feel more natural to me.

Simn:
Regarding `from this`, you can always do something like this:

```haxe
private typedef This = MoreComplicated<Type -> With, Parameters>
abstract Foo(This) from This
```

And I also think that `from to Int` looks like a syntax error...

vonagam:
Yes, i know about typedef, but shortcut is shorter.

I though about including other candidates for `from to` - `from-to`, `from and to`, `as` and so on. `from to` - was the clearest and without additional words so went with this for initial proposal (in drawbacks i wrote that for some it will not look like an ideal). I am ok with other variant being chosen, no resistance from me in that regard.

emugel:
Regarding `from` and `to`, the time spent writing the first abstract is not necessarily something that should be automated. 

Often you need to carefully think whether you really want to have both `to` and `from` because of problems it can create. If there is a short syntax, maybe the time spent typing can not be used to think and it creates a bad habit.

RealyUniqueName:
I think declaring a new abstract type is something happening frequently. 
Existing solution with `typedef` is enough. Especially in terms of reusability because `from` and `to` mean that you have that type used somewhere else in your code.

vonagam:
> you have that type used somewhere else in your code

Not necessary if you depend on structural unification or paramed function variance. Example of last:
```haxe
@:callable @:forward abstract Predicate<T>((value: T) -> Bool) from (value: T) -> Bool to (value: T) -> Bool {...}

// with typedef
private typedef This<T> = (value: T) -> Bool;
@:callable @:forward abstract Predicate<T>(This<T>) from This<T> to This<T> {...}

// with proposal
@:callable @:forward abstract Predicate<T>((value: T) -> Bool) from to this {...}
```

Some statistics - out of 34 abstracts in std that define at least one direct cast - 26 define `from to` cast to their underlying type.

vonagam:
Closing as this proposal has no chance of passing right now.
Sad that such common scenarios do not have more readable and concise syntax.

Simn:
I'll give you "concise", but I'm not so sure about the "readable" part, especially the `from to` suggestion.

I don't completely hate the `from this` idea, but I have some reservations about introducing `this` at type-level. However, if we go for #36, something like that is going to be required anyway.

vonagam:
Maybe `cast` can be used in place of `from to`.
As `from` and `to` mean "cast from" and "cast to", simple `cast` will mean "cast from and to".
```haxe
abstract Foo(Int) cast Int {}
```

kLabz:
Would a syntax like `abstract X(Y) from/to Y {}` be possible?

PR #80, opened 2020-08-07
Matcher: setting variables and guard as expression in a pattern

vonagam:
`pattern.set` for setting variables without capturing involved:
```haxe
case Some(int) | None.set(int = 0):
```
`pattern.where` for guards:
```haxe
case Left(int).where(int < 0) | Right(int).where(int > 0):
```
[Render.](https://github.com/vonagam/haxe-evolution/blob/matcher-additions/proposals/0000-matcher-additions.md)

This is already implemented on [this branch](https://github.com/vonagam/haxe/tree/pattern-set).

A possible alternative name for `set` - `vars`, for `where` - `with`.


Simn:

PR #81, opened 2020-08-14
Notation for native optional arguments

vonagam:
Proposed notation for native optional non-nullable arguments that should always be at the end of an argument list, cannot be skipped with nulls in bindings (or with exotic haxe's type skipping), cannot be null-padded, does not require providing a default value that may go out of sync with described extern, does work clearly with old function notation:

```haxe
?!Int->Void

function(?!value: Int): Void;
```

[Render.](https://github.com/vonagam/haxe-evolution/blob/native-optional/proposals/0000-native-optional.md)

Relevant issues: [3826](https://github.com/HaxeFoundation/haxe/issues/3826) [7925](https://github.com/HaxeFoundation/haxe/issues/7925)

Alternative syntax can be single postfix question mark instead of prefix one:
```haxe
Int?->Void

function(value?: Int): Void;
```

### DIFF
--- a/proposals/0000-native-optional.md
+++ b/proposals/0000-native-optional.md
@@ -1,0 +1,34 @@
+# Notation for native optional arguments
+
+* Proposal: [HXP-NNNN](NNNN-native-optional.md)
+* Author: [Dmitrii Maganov](https://github.com/vonagam)
+
+## Introduction
+
+Syntax and proper way to specify native optional arguments.
+
+## Motivation
+
+Current lack of clear distinction between haxe generated optional (nullable) arguments and native (non-nullable) optional ones creates bunch of problems for working with external code and even internally for null-safety.
+
+```haxe
+?Int->Void // Can you pass a null to such signature or not?
+
+function(value=0): Void; // What about that one?
+```
+
+## Detailed design
+
+Proposed notation for native optional non-nullable arguments that should always be at the end of an argument list, cannot be skipped with nulls in bindings (or with exotic haxe's type skipping), cannot be null-padded, does not require providing a default value that may go out of sync with described extern, does work clearly with old function notation:
+
+```haxe
+?!Int->Void
+
+function(?!value: Int): Void;
+```
+
+Ideally for such arguments targets should generate appropriate native not-nullable optional arguments.
+
+## Impact on existing code
+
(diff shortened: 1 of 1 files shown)


vonagam:
Alternative without additional notation - some short typedef. `Opt<T>` for example.

More drastic change of course is to go back to the original design and treat `?Int` argument as native non-nullable one. Which is cleaner/proper, but don't know about whenever it is considered possible at this point.

hughsando:
I prefer the 'inline' keyword and make it work across all functions on all targets as call-site inlining, eg:
```haxe
     function f(inline a:Int=1):Int;
```
the intention of the substitution is completely clear here - no problems with inheritance or overloading.

The optional-externs can already be accurately represented using overload:
```haxe
@:cppFileCode('
const char * stringDup(const char *ptr, int len=-1)
{
   if (len==-1)
      for(len=0;ptr[len];len++);
   char *result = new char[len+1];
   for(int i=0;i<len;i++)
      result[i] = ptr[i];
   result[len] = 0;
   return result;
}

')
class Test
{
   @:native("::stringDup")
   @:overload(function (s:cpp.ConstCharStar) : cpp.ConstCharStar{ })
   extern public static function stringDup(s:cpp.ConstCharStar, len:Int) : cpp.ConstCharStar;

   public static function main()
   {
      trace( stringDup("hello",4) );
      trace( stringDup("hello") );
   }
}
```
So this case is maybe not so important.  Also, the first 'inline' case should work too, but requires the "magic" -1 constant to be replicated, which just an unfortunate implementation detail.

vonagam:
How would such `inline` optional parameter be represented in a function notation?

> The optional-externs can already be accurately represented using overload

How do you override an external method with optional arguments?

hughsando:
For a full haxe solution, the functional type is going to need to contain the default, so something like:
```haxe
var func: (name:String, inline topScore:Int = 0)->Void = null;
```
I guess here,  you can redefine the default value or even apply it to a function that does not have a optional argument because the default is in the signature, not the function call.  Mostly though, type inference can copy this value.

You can't override native functions.  If you want to do that, you could do something like make a haxe function call the native function, and then override the haxe function.  Here, you would be using the haxe defaulting methods so existing logic would apply.

vonagam:
> You can't override native functions.

Why?

hughsando:
Perhaps post an example of what you are trying to achieve - it would be most helpful for discussions.

vonagam:
Opened #9840, depending on how it go will update (or close) the PR.

Thanks for `inline` proposal, it might become handy later.

PR #83, opened 2020-10-14
Allow enums to opt-out of publishing their constructors in the implicit global scope.

0b1kn00b:
There's a couple of design principles underlying this that I'll go into if there is any interest.

### DIFF
--- a/proposals/0013-selective-qualified-enum-access
+++ b/proposals/0013-selective-qualified-enum-access
@@ -1,0 +1,41 @@
+# ::Selective Qualified Enum Access::
+* Proposal: [HXP-0013](0013-selective-qualified-enum-access.md)
+* Author: [Haxe Developer](https://github.com/0b1kn00b)
+
+## Introduction
+
+Allow control over the publishing of Enum constructors in the global scope.
+
+## Motivation
+
+Namespacing is important because it allows natural and comprehensible names within any given scope. 
+
+Enum constructors break with this convention for the sake of brevity, but can cause problems when integrating several grammars and publishing artifacts. When publishing a programming library that requires the use of enums it's necessary to work around this flaw. 
+
+Given, that for simplicity sake, it's polite to require as few imports as possible; naming conventions must be applied so as not to cause aliasing, which nullifies the advantages of namespaces.
+
+Without those naming conventions, it's necessary to push enumerations into separate packages and then explain somewhere where the missing piece is.
+
+#### tldr:
+
+It should be possible to suggest that enum constructors be used qualified so as to respect the global scope.
+
+## Detailed design
+
+Using a metadata tag `@:qualified_enum_access` should cause unqualified references to it's constructors to be inaccessable. `Enum.Constructor` remains valid. 
+
+Possible override in the same manner of `@:access`: `@:unqualified_enum_access`
+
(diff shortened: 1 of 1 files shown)


proposals/0013-selective-qualified-enum-access.md
```diff
@@ -0,0 +1,41 @@
...
+
+Using a metadata tag `@:qualified_enum_access` should cause unqualified references to it's constructors to be inaccessable. `Enum.Constructor` remains valid. 
+
+Possible override in the same manner of `@:access`: `@:unqualified_enum_access`
+
+## Impact on existing code
+  None I can think of.
+
+## Drawbacks
+  It would make some pattern matching code more verbose if the override isn't implemented, although `import LongEnumName in L` isn't too bad.
```
jonasmalacofilho, line 33: This could be solved by allowing enum constructors to be explicitly imported into the local namespace using a wildcard:<sup>1</sup>

```haxe
import LongEnumName.*;
```

This way unqualified references would still be possible even for `@:qualified_enum_access` enums, and the constructors could even be exposed to entire particular packages/folders using `import.hx` files.

<sup>1</sup> Maybe also individually; even though that's less useful, it would be more consistent with importing other types.

---

P.S. It's basically how Rust handles enums: ~~constructors~~ variants are namespaced, but can be imported with wildcards; and commonly used variants are imported in preludes, of which the most notable is _the_ automatic Rust prelude that imports, among other things, the variants for `Option<T>` (`Some`, `None`) and `Result<T, E>` (`Ok`, `Err`).

0b1kn00b: I'm easy as to how it looks myself, so long as a measure of control can be had beyond shuffling imports and so-forth. like it.

0b1kn00b:
Say a real world example: socket protocol. Typically you want something you can put over the wire, so rather than an interface, you use nested enums to describe the various bits and you can wrap as you step up the abstraction layers. 
It looks to break down (atm) into local client, remote client, local server, remote server as functionality goes, each with a slightly different input set. 

Queue masses of faffing with conventions `LocalServerRequest` containing `LSR_Open`, `RemoteServerRequest` containing `RSR_Connect` or whatever. 

It works, but it's less than elegant, so that's really the basis.

Aurel300:
I have been in the past affected negatively by certain rules of the name resolution system. But I feel that this solution is a bit of a hacky workaround, when really the "proper" solution would be to rethink that system in the first place (addressing more than just enum constructors).

0b1kn00b:
I don't disagree with that necessarily, I'm just not sure there is any way of doing it without breaking stuff. 

My view of this here is that the workarounds are *nasty*, so I thought I would flag it

0b1kn00b:
I think in the Nom, which is a rust parser framework, there is an operator to clear the scope. Would that be a better way of looking at it?

Simn:
lean-reject: weak proposal, questionable benefit

Simn:

The proposal addresses a valid concern, but the way it aims to solve it doesn't seem optimal. We agree that it would be useful to "hide" enum constructors in some cases, but this would have to be designed in a broader manner. Also, there's concern that any such feature should be accessible on the import-site, not just in the type definition.

PR #84, opened 2020-10-22
Local static

YellowAfterlife:
This had been on my mind for a while - I think allowing to declare "situational" static variables that are not visible outside of method is a good tradeoff between convenience and implementation costs.

[Rendered version](https://github.com/YellowAfterlife/haxe-evolution/blob/local-static/proposals/0000-local-static.md)


Aurel300:
I have found myself kind of wanting this every now and then, and indeed, implementing it with macros for the important cases. On the other hand, this does introduce hidden state and makes reading code a bit less accessible to newcomers…

hughsando:
I like it.  Although it has hidden state, the state is more obvious than if you declare the variable further away from the function.
This pattern is pretty common in C.

Geokureli:
Would this mean that we could also have method scoped local inline vars, too or would that be a separate thing altogether?

Simn:

PR #85, opened 2020-11-22
Null coalescing operator

RblSb:
[Rendered version](https://github.com/RblSb/haxe-evolution/blob/null-coalescing/proposals/0000-null-coalescing.md)

### DIFF
--- a/proposals/0000-null-coalescing.md
+++ b/proposals/0000-null-coalescing.md
@@ -1,0 +1,58 @@
+# Null coalescing operator
+
+* Proposal: [HXP-0000](0000-null-coalescing.md)
+* Author: [RblSb](https://github.com/RblSb)
+
+## Introduction
+
+Provide syntax for null coalescing operator, which returns the left operand if it is not null, and otherwise returns the right operand.
+
+## Motivation
+
+This operator exists in many other languages and makes null checks simplier and more readable. I think this would be nice addiction to Haxe syntax and future null-safe traversal operator.
+
+## Detailed design
+
+Example with usage of this operator, when `options` object is read-only:
+```haxe
+public static function create(options:Options) {
+  final bgColor = options.bgColor ?? 0x000000;
+  final title = options.title ?? "Cool Game";
+  setOptions(bgColor, title);
+
+  setSize(options.size ?? {w: 20, h: 10});
+}
+```
+This would be same as:
+```haxe
+public static function create(options:Options) {
+  final bgColor = options.bgColor == null ? 0x000000 : options.bgColor;
+  final title = options.title == null ? "Cool Game" : options.title;
+  setOptions(bgColor, title);
+
+  setSize(options.size == null ? {w: 20, h: 10} : options.size);
+}
+```
+
+Another possible usage with future null-safe traversal operator, if it will be implemented:
+```haxe
+return options?.advanced?.id ?? 0;
+```
+
+Return type of null coalescing operator will be non-nullable, same as with conditional operator. Incorrect type of right operand should error as before. In case of usage with multiple times operands are executed in left to right order, if no brackets are used:
+```haxe
+trace(foo ?? bar ?? 5);
+// is ((foo ?? bar) ?? 5)
+```
+
+## Impact on existing code
+
+Should be no impact.
+
+## Alternatives
+
+`foo ?: bar` syntax for this feature, that is used in some languages instead. But in this case would be harder to detect ternary operator typos.
+
+## Opening possibilities
+
+## Unresolved questions


RealyUniqueName:
Null check + default value is quite common operation. 
And null coalescing would play nicely with null safety feature because of the typing:
```haxe
var s:Null<String> = getNullableString();
$type(s == null ? 'default value' : s); // Still `Null<String>`
$type(s ?? 'defaultValue'); // would yield `String`
```

emugel:
Seeing Haxe more and more as very high-level language, it would be a nice thing to have.

But I favour the `?:` variant for 2 reasons:

1. `?:` is very consistent with the ternary operator, making it quite logical.
2. to me `??` is typographically ugly, reminiscent of hastly written scripts (with `????!!!` all the place), distracting programmer. 

```haxe
function main() {
    var x : Null<Foo> = getFoo();
    var o = { foo : x ?= Foo.empty() };  
}
```

vs

```haxe
function main() {
    var x : Null<Foo> = getFoo();
    var o = { foo : x ?? Foo.empty() };  
}
```

The first variant jumps much less at your eyes IMO.

emugel:
Reading about `?:` on https://en.wikipedia.org/wiki/Elvis_operator
there is also the another alternative usage for the `?:` operator, like in C++ or PHP. 
*tl;dr*  it couldn't be useful in Haxe because in these languages the concept of evaluating to true is so much less restrictive.

>> The name "Elvis operator" refers to the fact that when its common notation, ?:, is viewed sideways, it resembles an emoticon of Elvis Presley with his quiff.[1]
>> 
>> A similar operator is the null coalescing operator, where the check for boolean truthiness is replaced with a check for non-null instead. This is usually written ??, and can be seen in languages like C# (or Typescript).

In short, the usage of `?:` in C++ and PHP is to repeat the preceding result (for instance `var x = f() ?: g()` is equivalent to `var x = f() ? f() : g()`, but without evaluating `f()` a second time). In C++ for example an non-empty string, non-zero int, or a non-null object would evaluate to true making this very useful.

In Haxe, since nothing can evaluate to `Bool` implicitely, there would be no such use; hence we could not use it at all, or use it for *null coalescing*. C# also has `??=`, which is overkill IMO.

Aurel300:
> C# also has `??=`, which is overkill IMO.

Well, if `??` becomes a binary operator, I don't see a reason to disallow it in `OpAssignOp` for the `??=` combination. It shouldn't have the short-circuiting issues of `||=` and `&&=`. Besides, I find it pretty common to need to assign a default value if something is `null`.

emugel:
Fair enough, but the point of null coalescing operator is to reduce the LOC and the number of intermediate variables; that is why I think it is not as useful. 

Besides, `p ??= Foo.empty();` is less clear than `if (p == null) p = Foo.empty();` I think,
or than `p = p ?: Foo.empty()`. 

Let's see:

```haxe
pWindow ??= Foo.empty();

if (pWindow == null) pWindow = Foo.empty();

pWindow = pWindow ?: Foo.empty();

pWindow = pWindow ?? Foo.empty();
```
But it's probably a matter of practice, and in the end maybe you are in the right.

RblSb:
I would like to have many things from Kotlin, but when wiki says
> This is an unusual choice of symbol, given that ?: is typically used for the Elvis operator, not null coalescing, but it was inspired by Groovy (programming language) where null is considered false.

I would stick on C#/JS/etc option. Even if it looks more strange - it's pretty common and easier to type. Also don't think that ternary operator and null coalescing are that similar, so we shouldn't unify it to same syntax.
`??=` seems interesting, but i'm not brave enough to request this feature for now, something like `this.x = y ?? this.x` looks good enough.

YellowAfterlife:
> C# also has ??=, which is overkill IMO
`??=` is perfect for non-constant default argument values, e.g.
```haxe
function filter(fn:T->Bool, ?out:Array<T>):Array<T> {
    out ??= [];
    for (item in items) if (fn(item)) out.push(item);
    return out;
}
```
allowing to both call the function as-is to allocate a one-off array and to call it with an existing array to concat items to the end of it.

`a = a ?? b` introduces a bit of redundancy in comparison as we are reassigning `a` (which can be more complex than a local variable) even if it was fine.

nickmain:
Another vote for "??" since that is what Swift uses.

lublak:
Especially the null-safe traversal operator would make a lot of sense when working with json files.

```hx
return options?.advanced?.id ?? 0;
```

vs

```hx
return options == null ? 0 : (options.advanced == null ? 0 : (options.advanced.id == null ? 0 : options.advanced.id));
```

or

```hx
return options == null || options.advanced == null || options.advanced.id == null ? 0 : options.advanced.id;
```

Simn:

PR #86, opened 2020-11-24
Enum abstract instances

markknol:
This proposal would allow instances as enum abstract values (generated as static singletons) which is currently not possible (error: `Inline variable initialization must be a constant value`)

### DIFF
--- a/proposals/0000-enum-abstract-instances.md
+++ b/proposals/0000-enum-abstract-instances.md
@@ -1,0 +1,135 @@
+# Enum abstract instances
+
+* Proposal: [HXP-NNNN](0000-enum-abstract-instances.md)
+* Author: [markknol](https://github.com/markknol)
+
+## Introduction
+
+Allow instances as enum abstract values. 
+
+```haxe
+@:forward
+enum abstract Person( { final name:String; final age:Int; } )  {
+  final Mark = { name: "Mark", age: 21 };
+  final John = { name: "John", age: 25 };
+  final Elisa = { name: "Elisa", age: 28 };
+}
+```
+This would allow to use the enum be used in switches too:
+```haxe
+final person:Person = Mark;
+
+var message = switch person {
+  case Mark: 'Hello from The Netherlands, my name is ${person.name}';
+  case John: 'Hello from the USA, I am ${person.age} years old';
+  case Elisa: 'Hello from Japan';
+}
+
+trace('Message: $message');
+```
+
+## Motivation
+
+Currently when you try to use instance values in enum abstract the compiler gives an error:  
+`Inline variable initialization must be a constant value`
+
+This can lead to using workarounds that's more error-phrone. Allowing instances could be a pragmatic and neat solution.
+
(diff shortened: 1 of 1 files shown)


RealyUniqueName:
Won't this make our ordinary enums obsolete?

RealyUniqueName:
I guess it's one more step to replace ordinary enums.
The step is to make enum abstracts parameterized :)
```haxe
enum abstract Person( { final name:String; final age:Int; } )  {
  final Mark = { name: "Mark", age: 21 };
  final John = { name: "John", age: 25 };
  final Elisa = { name: "Elisa", age: 28 };
  final Fifties(name:String) = { name:name, age:50 }
}
//...
switch person {
  case Fifties('Herman'): trace('Herman is 50 years old');
  case _:
}
```

emugel:
Some things are a bit confusing yet for me.
Something from your example:
```
enum abstract Person({ final name: String; final age:Int }) {
  final John = { name: "John Bon", age: 65 };
  final Jane = { name: "Jane Dane", age: 18 };
  final Mark = { name: "Mark Hark", age: 36 };
  final Elisa = { name: "Elisa Disa", age: 42 };
}
```
Does that mean only these 4 person belong to the enum abstract? 
I think it is supposed to be, right?

Does that mean a `from { final name:String, final age:Int }` clause is *forbidden*? 

Another question (it was tipped in your final questions), are non-final fields forbidden? 

```
enum abstract Person({ name: String, age:Int }) {
  final John = { name: "John Bon", age: 65 };
  final Jane = { name: "Jane Dane", age: 18 };
}

var p = John;
p.name = "Jane Dane";
p.age = 18;
switch p {
  case John:
  case Jane:
// Is it Jane or John? And if you set the age to 17?
}
```

Simn:
We didn't reach a conclusion on this proposal in the haxe-evolution meeting today.

There are some open questions here related to the identity and equality of such values. We agreed to postpone the decision here and discuss this at a later point.

EliteMasterEric:
Looking at this proposal, it seems that the core issue here is the inability to associate a single enumerated value with several data points. The syntax provided however seems complicated and confusing for this use case.

It may make more sense to extend the functionality of the bare `enum` type instead, to allow creating enum constructors which contain constant values.

```haxe
// The syntax for this probably needs a second opinion or refinement
enum Color(Int, Int, Int) {
  Rgb(r:Int, g:Int, b:Int);
  Red(255, 0, 0);
  Green(0, 255, 0);
  Blue(0, 0, 255);
  Cyan(0, 255, 255);
  Magenta(255, 0, 255);
  Yellow(255, 255, 0);
  DarkRed(r:Int, 0, 0);
}

class Main {
  static function main() {
    var color = getColor();
    switch (color) {
      case Red(r, g, b):
        trace("Color was red, full intensity " + r);
      case DarkRed(r, g, b):
        trace("Color was red, intensity " + r);
      case Rgb(r, g, b):
        trace("Color had a red value of " + r);
    }
  }
}
```

PR #90, opened 2021-10-06
Number Separator

MrcSnm:
Incredibly useful feature for better code reability,  simple to implement, breaks nothing.
[Rendered version](https://github.com/MrcSnm/haxe-evolution/blob/master/proposals/0013-number-separator.md)

### DIFF
--- a/proposals/0013-number-separator.md
+++ b/proposals/0013-number-separator.md
@@ -1,0 +1,56 @@
+# Number Separators
+
+* Proposal: [0013-number-separator](0013-number-separator.md)
+* Author: [Hipreme|Marcelo Silva Nascimento Mancini](https://github.com/MrcSnm)
+
+## Introduction
+
+Make the following syntax available: 1_000_000 as an alternative for 1000000
+
+## Motivation
+
+I believe that number separator is achievable with macros, although I'm
+pretty sure that it would unnecessarily increase compile time and it is
+something pretty simple to implement.
+
+D and Javascript are successful languages which contain those features,
+the make a lot easier to read when working with great numbers, as transforming
+from nanoseconds to seconds: 1 / 1000000000. It is fairly impossible to read
+that number and don't count the zeros to know which number is that, with separators,
+we can make triples: 1_000_000_000, then you will check that exists 3 triples of 0.
+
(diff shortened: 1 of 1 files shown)


Geokureli:
> I believe that needing to import a module, executing a string.replace and modifying the AST is a lot slower than jumping an index when a `_` is found within a number. Readability was one of the reasons for this proposal too, having to call a function to do that misses the point

I doubt it will ever have a significant impact on compile time. regardless, you're right about readability. not to mention macros seem to be the default response to language feature requests, even though macros are not very accessible to beginner/intermediate haxe devs

RealyUniqueName:
This would be very useful if Haxe had binary literals. But for decimals it doesn't seem so useful. 
What kind of tasks require a lot of big number literals all over the code base? And by "big" I mean at least six numbers before decimal point.
Usually you define such constant in a single place and refer to it by name, not by literal.

I think currently this feature would have a very limited usefulness. So, unless we get binary literals or unless this feature gets widespread in major languages I'd vote against this feature.

Gama11:
>unless this feature gets widespread in major languages I'd vote against this feature.

It actually _is_ widespread, to name a few:

- [Java 7 has `_`](https://docs.oracle.com/javase/8/docs/technotes/guides/language/underscores-literals.html#:~:text=In%20Java%20SE%207%20and,the%20readability%20of%20your%20code.)
- [Kotlin has `_`](https://github.com/Kotlin/KEEP/blob/master/proposals/underscores-in-numeric-literals.md)
- [C# 7 has `_`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-7.0/digit-separators)
- [C++ 14 has `'`](https://en.cppreference.com/w/cpp/language/integer_literal#Single_quote)
- [Python 3.6 has `_`](https://www.python.org/dev/peps/pep-0515/)

back2dos:
This is quite interesting for hex literals, where one could use it to tell the individual bytes apart more easily (in particular for ARGB/RGBA colors).

Simn:
I can see this being useful in some instances, but I also kinda hate how it looks...

Simn:

PR #91, opened 2021-10-11
Static Extension Meta Functions

lublak:
[Rendered version](https://github.com/HaxeFoundation/haxe-evolution/blob/4f21b84d269f2c0dfe5c92b9d338073e66ad2e52/proposals/0000-static-extension-meta-functions.md)


EliteMasterEric:
I think this proposal is clean and makes a lot of sense. I see a lot of potential for making syntax very convenient with this, I like `@:arrayAccess` on strings in particular and think it should be added to StringTools. :+1: from me.

Simn:
I can't see myself accepting any more implicit casts in the Haxe language, but I'm very much in favor of operator overloading via static extensions. This is a poorly written proposal though...

PR #92, opened 2021-10-13
haxe.Int64 Numeric Literal Suffix

Aidan63:
Would allow appending an integer literal with `i64` to create a `haxe.Int64`.

```haxe
final myInt64 = 800000000i64;
```

[Rendered version](https://github.com/Aidan63/haxe-evolution/blob/int64-literal-suffix/proposals/0000-int64-iteral-suffix.md)

### DIFF
--- a/proposals/0000-int64-iteral-suffix.md
+++ b/proposals/0000-int64-iteral-suffix.md
@@ -1,0 +1,42 @@
+# haxe.Int64 Numeric Literal Suffix
+
+* Proposal: [HXP-0000](0000-int64-iteral-suffix.md)
+* Author: [Aidan Lee](https://github.com/aidan63)
+
+## Introduction
+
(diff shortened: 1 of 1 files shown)


Simn:
Yes, this indeed looks better in every way.

haxiomic:
I wonder if we can do away with suffixes in favor of implicit typing – If the following worked (where long literals are implicitly typed as Int64) then I'm not convinced they'd be necessary

```haxe
var x: Int64 = 8589934592; 
var x: Int32 = 8589934592; // compiler error: literal is too long to be represented as Int32. Use cast to force and truncate 
var x = 8589934592; // long int literal automatically typed as Int64 
var x: Uint64 = 8589934592;
var x: Int64 = 123;
```

TheDrawingCoder-Gamer:
Imo Int64 is not good for regular use case because it's considered a class on some targets, which precludes it from being assigned to vars (very frustratingly making me split my bitflags into two enum abstracts). It's good to have a suffix but Int64 needs fixing

Simn:

PR #97, opened 2022-05-27
Constructor `this.arg` syntax

RblSb:
```haxe
class People
  final name:String;
  final age:Int;
  function new(this.name, this.age) {}
}
```
[Rendered version](https://github.com/RblSb/haxe-evolution/blob/this_arg/proposals/0000-instance_args.md)

### DIFF
--- a/proposals/0000-instance_args.md
+++ b/proposals/0000-instance_args.md
@@ -1,0 +1,78 @@
+# Constructor instance arguments
+
+* Proposal: [HXP-NNNN](NNNN-filename.md)
+* Author: [RblSb](https://github.com/RblSb)
+
+## Introduction
+
+Provide `this.arg` syntax for assigning constructor arguments to instance fields.
+
+## Motivation
+
+This operation is very common in object-oriented programming and with this syntax data-classes should be easier to write, read and change.
+
+## Detailed design
+
+Lets see a typical class example, that is used to keep some data:
+```haxe
+class People
+  final name:String;
+  final age:Int;
+  function new(name:String, age:Int) {
+    this.name = name;
+    this.age = age;
+  }
+}
+```
+This is how it would look with the new syntax:
+```haxe
+class People
+  final name:String;
+  final age:Int;
+  function new(this.name, this.age) {}
+}
+```
+With the second example there is less code duplication and no requirement to specify the types, they are derived from respective fields. `this.arguments` are set before all code in the constructor body.
+Access to `name` and `this.name` will be identical and mean access to the field `name`, like there isn't any argument variable in the constructor:
+```haxe
+class People extends Bug {
+  final name:String;
+  final age:Int;
+  function new(this.name, this.age) {
+    // after fields are assigned, constructor code starts execution
+
+    age++; // changes this.age
+
+    super();
+
+    name = 'Cooler $name'; // changes this.name
+  }
+}
+```
+This should mean less variable shadowing mistakes.
+
+`this.` arguments can be used with normal arguments in any order:
+```haxe
+function new(type:Int, this.context, isBlock = false) {}
+```
+
+If `foo` field for `this.foo` argument doesn't exist in a class, there should be a compiler error.
+
+This syntax is disabled outside of constructor functions.
+
+## Impact on existing code
+
+Should be no impact.
+
+## Alternatives
+
+There is a possible macro code generation replacement, but I think this syntax is simple enough to understand, to be in language itself.
+
+## Opening possibilities
+
+We can omit constructor body requirement, if there is already any `this.argument` in `new` args:
+
+`function new(this.name, this.age);`
+
+
+## Unresolved questions


nadako:
I like it and I think we actually discussed this idea on the dev Slack before, but I struggle to remember if we had any unclarities or something against it...

nadako:
I just found a discussion that this actually comes from Dart language, and also a link to a macro I wrote to implement similar thing: https://gist.github.com/nadako/273497023683e20f964ccd6ac0643422

Aurel300:
How do we express this in the AST? An extra (nullable) flag on [`haxe.macro.FunctionArg`](https://api.haxe.org/haxe/macro/FunctionArg.html) that can only be used on constructors?

RblSb:
I like `new final foo:Float`, this is less flexible, so you cannot change order of fields independent of argument order, but at the other side arguments order is easier to follow. But i think this needs something in constructor, to get attention for existed new fields, like:
`function new(this.args, bar:String) {}`
So you also can add non-field args before or after.

This also opens question, if we should assign `null` from argument to field with default value:
`new final count = 0;`

nanjizal:
```@:structInit``` is awkward to remember hard to google and verbose, and where I would like short cut most, I am still seeing repetition, it could be simpler.
```Haxe
class People_ {
    public function( public final this.name: String, private var this.age: Int ){}
}
abstract People( People_ ){
   public inline function child(){
       return this.age < 16;
   }
}
function main(){
    var boy: People = { name: Oliver, age: 12 };
    if( boy.child() == true ) trace( '$(boy.name) is child' );
}
```

nanjizal:
oops forgot the 'new'.. but may have over simplified from real haxe anyway.
You probably don't need 'this'... with my thought..  This should allow typedef construction by default as option.
```Haxe
class People_ {
    public function new( final name: String, public var age: Int ){}
}
```

nanjizal:
for lots of properties splitting by line would work fine.
```Haxe
class People_ {
    public function new( 
      final name: String
    , public var age: Int ){}
}
``

**Comment by rhynomatt** [comment]
@nanjizal I'm not a fan of this. it trades clarity for brevity, adding a totally different spot for something as fundamental as field declaration to go into. it could introduce sneaky fields if you ever want to mixmatch between regular declaration and in-constructor declaration:
```haxe
class Person {
  public var name: String;
  public var age: Int;
  public var friends: List<Person>;

  public function new(this.name, age: Int, ?friends: List<Person>, var id: Int) {
    this.age = age;
    this.friends = friends ?? new List<Person>();
  }
}
```

player-03:
> But there is only one constructor. So one has to go back and forth between that and declarations to see which ones are initialized in it.

The thing is, that's true with or without the new syntax.

```haxe
class Airplane {
  var captain:Pilot;
  var copilot:Pilot;
  // ... and so on ...
  public function new(captain:Pilot) {
    this.captain = captain;
  }
}
```

`captain` is initialized, `copilot` isn't, and you have to scroll down to find that out. But there's also already a couple solutions available:

```haxe
class Airplane {
  final captain:Pilot;
  var copilot:Null<Pilot>;
  // ... and so on ...
}
```

Now you know that `captain` must be set, and you should expect `copilot` to be null sometimes.

Alternatively, just write comments. I'd argue that if you're writing a class that's big enough for this to be a problem, you can take responsibility for code clarity. Haxe doesn't need to force it, at least not at the expense of constructor usability. (More on that later.)

> I'm not a fan of this. it trades clarity for brevity, adding a totally different spot for something as fundamental as field declaration to go into.

Agreed. If I was searching for a certain field, I would definitely not think to check the constructor.

I do want shorter code, but not at the expense of clarity.

---

So at this point I've spoken out both for and against clarity. Evidently this isn't a simple issue. Let's talk about that.

I see this as a tug-of-war between brevity and clarity/usability.

- **Brevity** is the whole point of this proposal. Existing code is longer than it needs to be, and we'd like to fix that with some syntax sugar.
  - The original proposal saves one line per variable (`this.varName = varName`), while keeping the constructor arguments about the same (you have to add `this.` but can remove the type hint).
  - The tink_lang approach removes even more code. You no longer need to type out the constructor arguments (beyond a single `_` for all of them). You do have to mark the field declarations in some way (possibly with `= _` or `new final`).
  - The "declare vars in constructor args" approach produces the shortest code. You only need to type each variable once, without so much as an extra `_`.
- **Clarity and usability** come in many forms, and depend on use case.
  - Shorter code is often clearer and easier to write. That means the "constructor args" approach has a built-in advantage.
  - [As rhynomatt described above](#issuecomment-1184656502), if you mix regular declaration and constructor args declaration, suddenly the code becomes a lot less clear. Still easy to write, noticeably harder to read.
  - [As back2dos described](#issuecomment-1150912404), if you only have regular declarations, then you need to scroll down to the constructor to see which ones are initialized. Whereas if you mark them with `= _` or `new final`, then you can instantly see which is which.
  - If you take the tink_lang approach and include a placeholder in the constructor arguments, then the constructor becomes less clear _and_ less usable. Some users straight-up won't know what `function new(foo:String, _, bar:String)` means, and those who do will still have to scroll back up to the variable declarations to find which arguments to pass. (Hope they don't miss any!)
  - Speaking of the constructor signature, `function new(this.name, this.age)` doesn't tell you the variable types. `name` is easy enough to guess, but for `age` you'll have to scroll back up to check if it's an `Int` or `Float`. (Assuming for the sake of example that this is a large class.)
  - Users may want to write separate documentation for each instance variable, but if the variable is declared as a constructor argument, you'll have really limited access to formatting. Argument documentation isn't allowed to use line breaks, which in turn means you can't do paragraphs, lists, tables, or code blocks.

---

Aside: code completion (and related tools) can improve clarity/usability. However, it isn't always available, so I'd argue we shouldn't rely too heavily on it.

- Code completion can tell you which arguments the constructor takes, even if the constructor definition omits type hints (like the original proposal) or includes a placeholder argument (like the tink_lang approach).
- vshaxe could add an [inlay hint](https://github.com/vshaxe/vshaxe/wiki/Inlay-Hints) to each variable declaration, so that you can see which are assigned via constructor argument without scrolling down.

player-03:
My opinion: I prefer RblSb's original proposal over the others, but I find it hard to read and I'm not ready to endorse it.

I feel like I'd have trouble justifying _any_ of the proposals to a newbie Haxe programmer. They're all hard to decipher at first glance. Why are some of these arguments prefixed with `this.` and then never referenced? Why do some of the arguments say `var` and `final`, and are also never referenced? What are all these underscores?

Given that we already have `@:structInit`, I'm starting to think that's all we need. At the very least, it's easier to Google than any of these other options.

RblSb:
I think this suggested alternative is cleaner:
```haxe
class Airplane {
	new var captain:Pilot;
	new var copilot:Pilot;
	new var thrust = .0;

	public function new(this.args, isFast:Bool) {
		if (isFast) captain.name = "Flash";
	}
}
```
But this adds two concepts, `new var` and `this.args` or some other way to mark field args position in constructor.
So idk if current or this idea can be accepted. Both options will simplify refactoring, because even renaming field in IDE will not rename constructor argument currently, but `new(this.foo)` for example can be referenced as field to rename it.

PR #98, opened 2022-07-05
Traits

TheDrawingCoder-Gamer:
This proposal adds  traits, which are interfaces that have default implementations. 
It also adds implementations for traits, which can be applied anywhere.

[Rendered Proposal](https://github.com/TheDrawingCoder-Gamer/haxe-evolution/blob/traits/proposals/0000-traits.md)

### DIFF
--- a/proposals/0000-traits.md
+++ b/proposals/0000-traits.md
@@ -6,6 +6,7 @@
 ## Introduction
 
 Introduces a new top-level declaration type to haxe that acts like an interface, but allows default methods to be added. 
+It also allows one to implement these traits anywhere on any class.
 
 ## Motivation
 
@@ -14,7 +15,10 @@ default implementations in traits and interfaces respectively. Given that defaul
 implementations on interfaces have been rejected, this is the only remaining way 
 that they can be implemented in the compiler. Currently, there are macros that allow
 partial implementations, however they aren't, in my opinion, as good as a natural implementation, 
-like Java or Rust.
+like Java or Rust. 
+
+Taking heavily from Rust, this also allows one to define new methods on a class anywhere, and to allow truly generic functions
+to exist that actually do useful things. 
 
 ## Detailed design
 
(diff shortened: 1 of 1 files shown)


TheDrawingCoder-Gamer:
Adding Impl would upheave a lot of syntax... however, haskell and rust both allow defining instances in either:
1. The module where the trait was defined
2. The module where the type being implemented was defined
Perhaps it would be useful to allow writing implementations. I'll try adding some of that spice.

TheDrawingCoder-Gamer:
Hm, perhaps for JVM and C#, a `@:native` metadata will generate traits as interfaces and it doesn't allow impls, so I will be double happy.

TheDrawingCoder-Gamer:
I was recently thinking about how haxe, in it's current state, would implement type classes. I drew a complete blank, which is not very good, as type classes allow for functional programming. Traits in rust act like type classes, and haskell has typeclasses. It allows for powerful generic functions than can be implemented after the fact.

Aurel300:
Traits are a powerful language feature, yes. They would also require massive changes to the language and the compiler to accommodate them, assuming they can even be encoded into all of our target languages without too much overhead. For example, the mangling-based encoding you describe works after all types and traits are known. This should be true when code is being generated but what about in the middle of the typing process? When the typer sees the declaration `class Foo implements X`, it is known that `X` is implemented, but what if `implements X for Foo` is in a separate module? Typing of the module (or its dependencies) might depend on a trait being implemented. (Rust solves this kind of problem with [coherence rules](https://rust-lang.github.io/chalk/book/clauses/coherence.html).) What about blanket implementations, i.e. implementing a trait if another trait is implemented on any type, or even on any type parameter of a concrete type?

Simply put, this might be too large to implement without reconsidering a lot of Haxe in the process.

0b1kn00b:
There's a paper from 2004 by Bart Groot called "Implementing Traits In Java" that I think deals with some of these issues. He goes into "the flattening property" which might allow you to present traits within the current AST with a little metadata. Section 3.4 "typing traits"  goes into this in detail. There's already a little juju to implement `abstract`, I don't think this would be too different, especially now abstract classes are there.

TheDrawingCoder-Gamer:
I don't think this represents what a "trait" is. Someone can later make a PR about TYPECLASSES not traits.

PR #99, opened 2022-10-01
Caller tracking.

Apprentice-Alchemist:
A replacement for that `?pos:haxe.PosInfos` magic.

[Rendered proposal](https://github.com/Apprentice-Alchemist/haxe-evolution/blob/proposal/track-caller/proposals/0000-track-caller.md)

### DIFF
--- a/proposals/0000-track-caller.md
+++ b/proposals/0000-track-caller.md
@@ -27,11 +27,11 @@ function foo(caller:haxe.PosInfos) {
 	trace(caller);
 }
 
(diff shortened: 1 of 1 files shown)


Simn:
You correctly point that that there's a problem with `PosInfos` handling vs. rest arguments, but then your design doesn't explain how to actually resolve that in the context of your proposal.

I don't hate the actual design, but I don't think this resolves the problem you want to resolve.

PR #100, opened 2023-01-06
Trailing Block Expressions

RobertBorghese:
Pass block expressions to macro functions like this:
```haxe
myMacroFunc(123) {
    // block expr
}
```

[Rendered version](https://github.com/RobertBorghese/haxe-evolution/blob/master/proposals/0000-trailing-block-expr.md)

### DIFF
--- a/proposals/0000-trailing-block-expr.md
+++ b/proposals/0000-trailing-block-expr.md
@@ -1,0 +1,152 @@
+# Trailing Block Expressions
+
+* Proposal: [HXP-NNNN](NNNN-filename.md)
+* Author: [Robert Borghese](https://github.com/RobertBorghese)
+
+## Introduction
+
+Allow block expressions to be appended to macro calls as final argument.
+```haxe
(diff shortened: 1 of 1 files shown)


logo4poop:
Why does it have to be a macro function, couldn't it be expanded for any lambda function?

RobertBorghese:
Could you clarify what you mean? Do you mean a [Kotlin-like feature](https://kotlinlang.org/docs/lambdas.html#passing-trailing-lambdas) of allowing trailing blocks to be treated as lambdas that are passed as the final argument to a normal function?

logo4poop:
Correct

Pign:
I think this makes it difficult to read in most cases and even creates ambiguous reading.

In your example, the "simple" way of reading the last line (without reading the first one) would be that myMacroFunc has a signature comparable to Int->(TrailingExpr->Void)->Expr. Which is exactly what your first line means but wouldn't it be possible for MyMacroFunc to have the following signature : Int->(TrailingExpr->Expr) ?
This makes it difficult to read and understand the code because of ambiguity.

PR #101, opened 2023-01-06
Integer data type

flashultra:
Sometimes it is not so easy to work  with  Int64 types in  Haxe. Recently anounced  literal suffix and number separator make things better , but real support for Int64 ( replacing abstract class with the native for the target) will  give better performace and will lead to more clean code. 
For example 
```haxe
  var v:Int64 = 9223372036854775807;
```
and not ```var v:Int64 = Int64.make(0x7FFFFFFF,0xFFFFFFFF);```
Well with nigth build it could be ``` var v:Int64 = 0x7FFFFFFFFFFFFFFF_i64``` , but still to be good to have option without suffix.
Other things could be adding universal UIn64 and unification of Int and Int32 types.

### DIFF
--- a/proposals/0000-integer-data-types.md
+++ b/proposals/0000-integer-data-types.md
@@ -1,0 +1,60 @@
+# Integer data types
+
+* Proposal: [HXP-NNNN](NNNN-filename.md)
+* Author: [MJ](https://github.com/flashultra)
+
+## Introduction
+
+Allowing easy assignment of Int64 type for all Haxe targets. Replacing with native Int64  , where is it possible and use abstract for others.
+
+## Motivation
+
+Before Int64 literal suffix, it was more inconvenient to use Int64 (by calling Int64.make() ) and to create Int64 constant.
+Still, sometimes is less readable when for example have this 
+```haxe
+    var x = 2147483i64;  // this is 2147483 ( Int64)
+	var y = 2147483164;  // this is 2147483164 ( Int32)
+```
+Using a number separator makes things better, but still be better to use ```var y:Int64 = 4294967296;``` vs ```var y:Int64 = 4294967296_i64;``` as we know the type of the variable.
+
+## Detailed design
+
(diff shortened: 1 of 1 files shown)


Apprentice-Alchemist:
> but real support for Int64 ( replacing abstract class with the native for the target) will give better performace and will lead to more clean code

That already happen on supported targets (C++, HL, C#, etc)
Using eg Float on JS might be possible, but we need to be careful to keep behavior consistent between targets. 

I'm also not a fan of the compiler implicitly deciding of the integer type.
Numbers without suffixes (and with no type hints) should default to Int and give an error for too large literals.
So `var x:Int64 = 0xFFFFFFFFFF;` should be allowed, but not `var x = 0xFFFFFFFFFF;`.

> Other things could be adding universal UIn64 and unification of Int and Int32 types.

This would be nice, but again, we need to be careful about target-specific behavior.

flashultra:
In practical is very rarely to use Int above 2147483647, so to not loose performance the current behaviour could be kept i.e. default Int (according to the target) and if someone want consistent overflow behaviour to use Int32.
I did some tests on that  ( for Firefox: https://www.measurethat.net/Benchmarks/Show/22863/0/int-32-bit-test#latest_results_block )
For Chrome results are better.
So, to not lose performance the current solution (Int and Int32) is maybe good.

Int64  for Javascript could be a tricky one. Using Bigint for replacement could lead to loose of performance. I did some test for : Int64 (Haxe) , BigInt (JS) and Float ( Haxe) . The float give wrong result. BigInt give better performance than the current Int64 abstract class. Here is the example: https://try.haxe.org/#3330D013
 
 |          | Int64 | BigInt | Float    |
 |-------|-------|--------|---------|
 | Time  |0.04929|0.00570 |0.00190  | 
 |Result |Correct|Correct  |Incorrect |

>Numbers without suffixes (and with no type hints) should default to Int 

Default Int looks as a good solution . The other one could be to to use only signed integer values i.e Int < Int64 < BigInt depending of the value.

Adding UIn64 will be nice addition. Some target already have a native one ( C/C++ - unsigned long long ; C# - ulong; ) for other could be emulated or used BigInteger.

PR #105, opened 2023-03-16
Package alias

hydroper:
A little extra syntax proposal

[Rendered](https://github.com/hydroper/haxe-evolution/blob/master/proposals/0000-package-alias.md#package-alias)

### DIFF
--- a/proposals/0000-package-alias.md
+++ b/proposals/0000-package-alias.md
@@ -17,10 +17,6 @@ instead of a lexical reference can be useful for giving an idea of what a refere
 
 ## Detailed design
 
-Describe the proposed design in details the way language user can understand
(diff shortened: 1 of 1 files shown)


Simn:
The `.*` confuses me here, my intuition would be `import ecma262.temporal as Temporal;`.

hydroper:
@Simn In ActionScript packages are not to be mixed with different symbols in name resolution, so that's one reason I see for `.*`.

fy0:
Can't agree more.

EliteMasterEric:
@hydroper Did you delete this PR by accident? Can you restore it?

linuxmetro:
@EliteMasterEric Hi, I've broken my phone chip sometime ago and lost access to @hydroper :(

PR #110, opened 2023-04-29
The `throws` statement

SomeoneSom:
[Rendered proposal](https://github.com/SomeoneSom/throws-keyword/blob/master/proposals/0000-throws-statement.md)


Simn:
I've always found this annoying to deal with in Java, but I suppose that comes from my rather non-corporate attitude. I don't like having to tell a compiler twice that I'm throwing some exception, once to throw it and then once to confirm that I actually mean it. So this doesn't seam very appealing to me, but I won't let that affect my judgement on this too much.

grepsuzette:
I share your point of view on why it's useful, but don't think we should add a syntax.
Most people won't use it, also having `function foo() :Void throws Class` complicates the signature.

I have been using `@:throws(Class) function foo():Void` for years.
For me it's better because:
- it feels more optional and less java-esque.
- no discomfort about how long it can get, because you can always put the `@:throws` on a separate line

tl;dr: I don't see the need of a more specific syntax.

back2dos:
Not an overly big fan of relying on exceptions so strongly, but this feels somewhat incomplete:

- Are the exceptions thrown part of the function signature? Are two functions with the same args and return type but differenent exceptions thrown the same or different function types?
- If a function returns something like Rust's `Result`, the caller is forced to deal with the error. Should the compiler enforce catching at some point? If not, what does this effectively accomplish?

SomeoneSom:
> Are two functions with the same args and return type but differenent exceptions thrown the same or different function types?

That's a good point, making this work with function types would require some strange syntax, I am not actually sure how that would be achieved. `(args) -> return -> exception` could potentially work, but this seems more like a hack than an actual solution.

> I have been using @:throws(Class) function foo():Void for years (on my personal, non-java, projects).

I never actually considered `@:throws` since it only actually does anything on the Java target, but this idea might be better suited as a new compiler meta tag (this could also work by just adding more stuff to `@:throws`, but then it performs different things on different targets, and moving the current `@:throws` to something like `@:javaThrows` would be a breaking change, though I have no clue how much it would actually effect).

SomeoneSom:
Closing as considering this, this would be better as compiler metadata.

PR #111, opened 2023-05-17
Typed metadata (again!)

SomeRanDev:
Another typed metadata proposal.

Similar to previous proposal with relevant discussion [here](https://github.com/HaxeFoundation/haxe-evolution/pull/73). Relevant wiki content [here](https://github.com/HaxeFoundation/haxe-evolution/wiki/Ideas#typed-metadata). Based off of @nadako's [comment](https://github.com/HaxeFoundation/haxe-evolution/pull/73#issuecomment-634289234) which seems to be the most popular idea.

```haxe
// MyMeta.hx
@:metadata function author(name: String): haxe.macro.Expr.TypeDefinition;

// Sleigh.hx
@:MyMeta.author("Santa")
class Sleigh {
    // ...
}
```

[Rendered proposal](https://github.com/RobertBorghese/haxe-evolution/blob/typed-metadata/proposals/0000-typed-metadata.md)

### DIFF
--- a/proposals/0000-typed-metadata.md
+++ b/proposals/0000-typed-metadata.md
@@ -82,7 +82,7 @@ There's a lot to cover here. A table has been provided for your convenience:
 
 | Topic | Description |
 | --- | --- |
-| Basic Rules            | The basic syntax and rules for a metadata "function" declaration. |
+| [Basic Rules](/0000-typed-metadata.md#basic-rules)           | The basic syntax and rules for a metadata "function" declaration. |
 | Haxe API Changes       | The changes to the Haxe API required. |
 | New Metadata           | List of new metadata used to configure metadata functions. |
 | Allowed Argument Types | List of argument types allowed for a typed metadata. |
--- a/proposals/0000-typed-metadata.md
+++ b/proposals/0000-typed-metadata.md
@@ -8,8 +8,7 @@
 
 # Introduction
 
-A typing system for Haxe metadata to validate a its arguments
-and optionally provide compile-time transformations.
+A typing system for Haxe metadata that can validate its arguments and optionally provide compile-time transformations.
 
 ```haxe
 // -----------------------
@@ -57,18 +56,23 @@ class MyClass {
 
 # Motivation
 
-Metadata can be a little tedious to work with sometimes.
+Sometimes metadata can be a little tedious to work with.
 
-When making a program involving Haxe metadata, it's simple to check for
-a specific meta's existance; however, things become way more complicated
-when working with its arguments. A good amount of boilerplate needs
-to be written to check if an argument exists, if it's the desired type,
-and then to convert the `Expr` object into the desired format.
+<ins>Making</ins><br/>
+When writing code using Haxe metadata, it's simple to check for
+a specific metadata's name; however, the arguments are a nightmare.
+A lot of boilerplate needs to be written to:
+ * check if an argument exists
+ * check if it's the desired type
+ * convert from `Expr` to a usable data type
 
-On the other hand using a Haxe library or framework that processes
-metadata can become troublesome, as there is no guarentee the metadata
-is documented properly. Not to mention there is no scoping control with
-metadata, so there may be naming conflicts.
+<ins>Using</ins><br/>
+On the other hand, using someone's Haxe code that processes
+metadata can become troublesome. There is no guarentee the metadata
+is documented properly, and there is no scoping control to prevent
+naming conflicts.
+
+- - -
 
 Typing metadata using function declarations provides a better format
 for finding, documenting, and error checking Haxe metadata.
@@ -86,15 +90,17 @@ There's a lot to cover here. A table has been provided for your convenience:
 | [Haxe API Changes](0000-typed-metadata.md#haxe-api-changes)             | The changes to the Haxe API required. |
 | [New Metadata ](0000-typed-metadata.md#new-metadata)                    | List of new metadata used to configure metadata functions. |
 | [Allowed Argument Types](0000-typed-metadata.md#allowed-argument-types) | List of argument types allowed for a typed metadata. |
-| [Allowed Return Types](0000-typed-metadata.m#allowed-return-types)      | List of return types allowed for a typed metadata. |
-| [Decorators](0000-typed-metadata.md#decorators)                         | The design of metadata that runs code from its function to modify its subject. |
+| [Allowed Return Types](0000-typed-metadata.md#allowed-return-types)      | List of return types allowed for a typed metadata. |
+| [Decorators](0000-typed-metadata.md#decorators)                         | The design of metadata that runs code from its function body. |
+
+&nbsp;
+&nbsp;
 
 ## Basic Rules
 
 A metadata can be declared using a function declaration with the `@:metadata` meta.
 
-Metadata functions are allowed to have no function body, though they can use one
-if desired. This will be covered later (see [Decorators](0000-typed-metadata.md#decorators)).
+Metadata functions are permitted to lack an implementation (similar to `extern` functions). Typed metadata with function code is still allowed and will be covered later (see [Decorators](0000-typed-metadata.md#decorators)).
 ```haxe
 @:metadata function myMeta(): Any;
 ```
@@ -123,33 +129,35 @@ Type parameters are not allowed on metadata functions.
 
 ### Metadata Scoping/Importing
 
-This metadata can be used on anything that allows metadata on it currently.
+Typed metadata can be used on anything that allows metadata on it currently.
 However, it follows the same scoping rules as Haxe functions. Meaning it must
-either be used with its full path (packages.Module.funcName) or be imported:
+use its full path or be imported:
 ```haxe
-@mypack.MyModule.myMeta function doThing() { ... }
-```
-```haxe
-import mypack.MyModule;
-@myMeta function doThing() { ... }
+@mypack.MyModule.myMeta
+function doThing() { ... }
+
+// OR
+
+import mypack.MyModule; 
+
+// static function: @MyModule.myMeta
+// module level:    @myMeta 
+
+@myMeta
+function doThing() { ... }
 ```
 
 ### Untyped Metadata
 
-If the Haxe compiler encounters a metadata entry it cannot type, how it is
-handled is an "Unresolved Question". 
+If the Haxe compiler encounters a metadata entry it cannot type, its behavior is currently an [Unresolved Question](0000-typed-metadata.md#unresolved-question). 
 
-For the time being, this proposal suggests throwing an error unless the
-define `-D allow-untyped-meta` is defined.
+For the time being, this proposal suggests throwing an error _unless_ `-D allow-untyped-meta` is defined.
 
 ### Metadata Target
 
-The return type of the metadata function declaration dictates where this metadata
-can be used. The `Any` type denotes a metadata can be used anywhere. Another example
-is `haxe.macro.Expr` restricts a metadata's usage to expressions.
+The return type of the metadata function declaration dictates where it's allowed to be used.
 
-A full list of allowed return types can be found at [Allowed Return Types](0000-typed-metadata.md#allowed-return-types).
-Any return type besides the ones listed there are not allowed and should result in an error.
+The `Any` type denotes a metadata can be used anywhere. The `haxe.macro.Expr` restricts a metadata's usage to expressions. A full list of allowed return types can be found at [Allowed Return Types](0000-typed-metadata.md#allowed-return-types). Any return type besides those are not allowed and should result in an error.
 
 ```haxe
 // use this anywhere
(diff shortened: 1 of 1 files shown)


proposals/0000-typed-metadata.md
```diff
@@ -0,0 +1,528 @@
...
+	compile-time (will not generate any rtti data).
+
+	Unresolved question: Should this force this meta to
+	be prefixed with a colon? There needs to be a way to signify
+	that when typing built-in Haxe metadata in the future.
+**/
+@:metadata
+@metadataCompileOnly
+@metadataRequire(@:metadata _)
+function metadataCompileOnly(): haxe.macro.Expr.Function;
```
kLabz, line 277: An alternative would be to use `@metadata function ...` to define a runtime metadata (`@foo`), and `@:metadata function ...` for a compile-time one (`@:foo`). Requiring to add `@metadataCompileOnly` to all compile-time metadata sounds very annoying (also, we have namespaces).

Drawback is that it's easy to make a mistake between `@:metadata function` and `@metadata function` but that should be caught very easily at use site.

kLabz: > also, we have namespaces

Btw, this seems like a limitation of this proposal to not handle metadata namespaces, which is unfortunate.

RobertBorghese:
> Why requiring the use of `Context.getDecoratorSubject()` instead of getting it as argument?

This was actually my original idea. Instead of `@:metadata`, you could have metadata functions work as static extensions, so any static function with `DecoratorSubject` as the first parameter is a metadata function. Then you could use `extern` to generate metadata functions without bodies.

The only reason I didn't commit to that in this proposal is I feel like body-less metadata functions are going to be used waaaay more frequently, so you end up writing much more to accommodate the less common use case? Plus nadako's comment seemed kinda popular. And finally, `Context.get____` feels like a standard way compile-time functions get information honestly (getBuildFields, getLocalClass, getLocalTVars, etc.).

But I'm glad you brought it up cause I thought that might be a better way too, and I'd be interested to hear what others think.

kLabz:
I don't know, really. All those switch + `case _: throw "Impossible";` etc seem pretty silly.
We could easily have a different signature for body-less metadata for those not to need to add a useless argument.

Speaking of body-less metadata, they seem pretty much useless as-is (unless we enforce typed metadata everywhere which I really doubt we'll do by default) as they don't even really ease reading their data from outside. For example I would assume we could somehow read something like `e.meta.author.name` from `@:author("Santa") blargh` expression with `@:metadata function author(name:String):Expr`.

proposals/0000-typed-metadata.md
RobertBorghese: > (edit: I guess it's in concurrence with fully qualified path there..)

Oh yeah, sorry if I didn't make it as clear in the proposal but you import like normal functions, otherwise you use full path like you would with a normal function call. I kind of touch on it a little [here](https://github.com/RobertBorghese/haxe-evolution/blob/ae1a4b9a9856e2f0653436973d60bdd64fc51cf5/proposals/0000-typed-metadata.md?plain=1#LL143C7-L143C7), demonstrating how using module-level vs static function changes how you'd write it with basic module import.

RobertBorghese: > Drawback is that it's easy to make a mistake between `@:metadata function` and `@metadata function` but that should be caught very easily at use site.

Yeah, this was certainly a tough thing to find an answer to. Maybe typed metadata should just support being used both ways. But if there's a colon, it doesn't generate rtti? Hmmm...

RobertBorghese:
That fair. 🤔

> something like `e.meta.author.name`

I like this a lot. I'll try and incorporate something like this into the proposal in a bit. `meta` already field on TypeDefinition/Field/etc, so how about something like `metaArgs`? (`td.metaArgs.author.name`)

Would need a way to handle having the same meta used multiple times... so maybe make it an Array in that case? `td.metaArgs.author[0].name`

kLabz:
I may be wrong but I remember reading something about the metadata function having some way to allow being used multiple times or not? That could potentially be used to allow either `.author.name` or `.related[0].link`. Typing there could be a bit annoying though; if so then I guess we could go with mandatory array access.

> meta already field on TypeDefinition/Field/etc, so how about something like metaArgs? (td.metaArgs.author.name)

We could have `x.meta` (current one) with raw data and `x.typedMeta` with typed metadata entries (allowing things like `td.typedMeta.author.name`)

Aurel300:
Merging typed metadata into the existing syntax seems unlikely to work out. In particular:

> For the time being, this proposal suggests throwing an error *unless* `-D allow-untyped-meta` is defined.

Does this mean that virtually all metadata everywhere would be rejected unless the flag is set, especially metadata coming from third-party libraries (which might not update at the same time as the compiler)? Not good…

Or is this saying, if the compiler fails to type a metadata, *when* a function for it exists and is resolved, it will complain? Also bad: a typo in the metadata name (or a forgotten import, etc) will simply be accepted silently.

Also: in your proposal you only use the runtime metadata syntax `@foo` and never `@:foo`—would there be a way to discern these? Is there any value in keeping metadata like this in RTTI?

IMO we need separate syntax for this. I think `@.` has been proposed before.

kLabz:
> IMO we need separate syntax for this. I think @. has been proposed before.

Also Simon proposed only allowing classes as metadata providers and disallowing "untyped" metadata starting with an uppercase letter. That handles the separate syntax issue.

RobertBorghese:
Updated proposal:
* [Reworked](https://github.com/RobertBorghese/haxe-evolution/blob/typed-metadata/proposals/0000-typed-metadata.md#untyped-metadata) untyped metadata section. Untyped metadata will only throw error if the module uses any typed metadata.
* [Removed](https://github.com/RobertBorghese/haxe-evolution/blob/typed-metadata/proposals/0000-typed-metadata.md#reading-arguments) `MetadataEntryTools`, use `typedMeta: StringMap<Dynamic>` on AST structures to extract arguments.
* [Added](https://github.com/RobertBorghese/haxe-evolution/blob/typed-metadata/proposals/0000-typed-metadata.md#allowed-argument-types) `Array<T>` and `Dynamic` to allowed argument types
* [Removed](https://github.com/RobertBorghese/haxe-evolution/blob/typed-metadata/proposals/0000-typed-metadata.md#metadata-arguments) extra metadata for configuring `@:metadata`, instead uses optional "options" argument:
```haxe
@:metadata({ allowMultiple: Bool }) function author(name: String): Any;
```
* Typed metadata that use `@:` must use `@:metadata({ compileTime: Bool })`. Tried to incorporate more examples of this.

proposals/0000-typed-metadata.md
```diff
@@ -0,0 +1,577 @@
...
+
+The following anonymous structure should be added to the `haxe/macro/Expr.hx` module:
+```haxe
+typedef FunctionPath = {
+	> TypePath,
+	functionName: String;
+};
+```
+
+This is a structure for storing type paths to functions. It is used as an argument type for metadata. Long story short, it allows for type paths that end with a lowercase identifier (`myFunc`, `Module.Sub.myFunc`).
```
kLabz, line 355: Why should that be function-specific? I'd expect it to be more like 
```haxe
typedef FieldPath = {
	> TypePath,
	field: String;
};
```

proposals/0000-typed-metadata.md
```diff
@@ -0,0 +1,577 @@
...
+// Unresolved question
+// Would it be possible to use Ref<Type.ClassField> instead?
+var ?field: Expr.Field;
+```
+
+&nbsp;
+
+### Reading Arguments
+
+There needs to be a mechanism for reading metadata arguments. To provide this, a new field `typedMeta: StringMap<Dynamic>` should be added to:
```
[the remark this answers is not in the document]
RobertBorghese: Ahh, sorry. Could you give a more clear idea? I thought you just meant a `Dynamic` originally. How would your system handle the scenario where there are two metadata with the same name? Like in the example: `Module1.myMeta` + `Module2.myMeta`?

proposals/0000-typed-metadata.md
```diff
@@ -147,11 +151,28 @@ import mypack.MyModule;
...
 ```
 
+&nbsp;
+
 ### Untyped Metadata
 
 If the Haxe compiler encounters a metadata entry it cannot type, its behavior is currently an [Unresolved Question](0000-typed-metadata.md#unresolved-question). 
 
-For the time being, this proposal suggests throwing an error _unless_ `-D allow-untyped-meta` is defined.
+For the time being, this proposal suggests printing an error for each metadata entry that could not be typed (no metadata function could be found for its name/path) ONLY IF within a module that meets the following conditions:
```
kLabz, line 160: I still don't think it's ok as default behavior.

proposals/0000-typed-metadata.md
RobertBorghese: That's a great point, oversight on my part.

RobertBorghese:
I favor the `@.` syntax. The reason it's not being proposed is because I'm trying to base this off the most 👍'ed comment on the previous proposal. Maybe I misread the general consensus cause I thought that having a new syntax seemed kind of unpopular (hence why it didn't pass), I would love to just write that and have it solve all the typed vs untyped compatibility issues.

proposals/0000-typed-metadata.md
RobertBorghese: Yeahhh, idk. This is mostly meant to be unresolved until the main question of whether `@.` or `@CapitalLetter` or some sort of new indicator should be used. Idk, like with a lot of things not married to it, just trying to throw out solutions. 🤔

proposals/0000-typed-metadata.md
```diff
@@ -0,0 +1,577 @@
...
+    ?allowMultiple: Bool,
+    ?compileTime: Bool,
+    ?platforms: Array<String>
+}): haxe.macro.Expr.Function;
+```
+
+| Argument Name | Default Value | Description |
+| --- | --- | --- |
+| allowMultiple | `false` | If `true`, this metadata can be used on the same subject multiple times. |
+| compileTime | `false` | If `true`, this metadata must be prefixed with a colon and does not generate rtti. |
```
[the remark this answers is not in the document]
RobertBorghese: I'm down for that.

wartman:
What if type checking metadata worked sorta like null safety does? That might give the most flexibility for backwards compatibility without needing to introduce all kinds of new syntax. Make it opt-in per package or even per class, basically.

```haxe
package my.package;

@:metadata({ rtti: true }) function foo(arg:String):Any;

// Could be something like this (at a class-level) or `--macro metadataTypeChecking('my.package', Strict)` (as an
// initialization macro). `Strict` could throw errors, `Warn` could just warn, `Off` could turn it off completely?
@:metadata.typeChecking(Strict)
class Foo {
  @foo('ok') final foo:String; // Fine
  @bar final bar:String; // Throws an error

  // etc.
}
```

Alternatively this could be opt-*out*, allowing you to turn off type-checking for various packages (but even then maybe only emit a warning by default?).

This probably won't work too well with the more ambitious parts of this proposal, but if all we really want is some completion and error checking, maybe it's enough? Heck, even if we go with a new syntax like `@.` having some way to check **all** metadata seems like it would be nice.

Simn:
Some remarks:

## Metadata targets as return types

This is a bit too clever. The main problem I see is that we cannot express the targets of something like `@:final`, which currently has `"targets": ["TClass", "TClassField"]`. It also looks somewhat alienating to deal with types from `haxe.macro.Expr` here. It would be better to just add the targets as an argument like a normal person would.

## `@:metadata` bootstrap

I guess this is mostly a note to myself for when I start implementing this, but `@:metadata function metadata` looks like a fun time. This has some implications for how we handle typer passes, because we can't really load the metadata function while typing the metadata function. 

## Decorators

Ok listen... I can see why you would want that, but it's a bit too much for a typed metadata proposal. At least at the first step. Metadata is (supposed to be) something very declarative, and executing macros for each of their occurrences is not. I would forego this entire part for now.

## Typed metadata activated like null-safety

I find @wartman's idea quite interesting. This might a good way to approach the issue indeed, especially because it allows "upgrading" untyped metadata to typed metadata. I'm not sure yet how this should be brought into scope (I think it should be at file/directory-level, probably utilizing import.hx), but my initial reaction that this is definitely the way to go.

One problem is that for some metadata, typing is required. I've been dealing with how our `@:strict` currently works recently, and it does some transformation by typing its arguments and then adding a modified `@:meta`. This means that some kind of metadata will always require typing, which should be considered.

This would mean that we _always_ try to find a metadata definition, and whether or not we error out if there isn't one depends on the mode we're in. I'm not sure if we really need a mode where we find a definition, but accept that it doesn't type properly. I'd estimate that this would mostly come from metadata naming conflicts.

0b1kn00b:
I've got something of an idea for typed metadata. If you look at the `haxe.macro.Type` and `haxe.macro.Expr` packages, everything boils down to Objects and Enums, or products or coproducts.

You could embed a Haxe representation of [EDN](https://github.com/edn-format/edn?tab=readme-ov-file#tagged-elements)  into the metadata declarations and still have both expressive power and typedness. or something like `JSONAst`, although [tagged elements](https://github.com/edn-format/edn?tab=readme-ov-file#tagged-elements) would allow constructors at macro time.


```haxe
enum MetadataItemType{
  TMEmpty;
  TMScalar;
  TMArray;
  TMObject;
}
enum MetadataItem{
 MEmpty;
 MScalar(c:haxe.macro.Expr.Constant);//perhaps without CIdent?
 MArray(array:Array<MetadataItem>);
 MObject(object:Array<StringKeyValue<MetadataItem>>);
 MApply(name:String,args:MetadataItem); 
}
```

mapping of metadata:  
```haxe
@:notifier("an argument")   => MApply("notifier",MEmpty)
@:more(oh_no,2) => MApply("more",MApply("oh_no",MScalar(CInt(2))))
@:value(true) => MApply("value",MScalar(CBool(true)));
@:path.into(true) MApply("path",MApply("into",MScalar(CBool(true))));
```
An abstract could allow `haxe.DynamicAccess` type access to Scalars.

```haxe

typedef StringKeyValue<T> = {
  public var key    : String;
  public var value  : T;
}
abstract MetadataAccess(MetadataItem) from MetadataItem{  
  @:op(A.B)
  public function get(key:String):Null<MetadataAccess>{
    return switch(this){
      case MObject(object) :
        var result = null;
        for(x in object){
          if(x.key == key){
            result = x.value;
            break;
          }
        }
        result;
      default : null;
    }
  }
  @:arrayAccess
  public function at(idx:Int):Null<MetadataItem>{
    return switch(this){
      case MArray(array) : array[idx];
      default : null;
    }
  }
  public function value():MetadataItem{
    return this;
  }
  public function ifScalar():Null<Dynamic>{
    return switch(__hx_metadata_type){
      case TMScalar  : this;
      default        : null;
    }
  }
  public var __hx_metadata_type(get,never):MetadataItemType;
  private function get___hx_metadata_type(){
    return switch(Type.typeof(this)){
      case TInt,TFloat,TBool,TClass(String) : TMScalar;
      case TClass(Array)                    : TMArray;
      case TObject | TClass(_)              : TMObject;
      default : throw 'bad type of MetadataAccess';
    }
  }
}
```
Raw identifiers could be represented as tagged elements, 
To type the metadata an interface with functions such:

```haxe
interface Hydrate{
 //if there are any bare identifiers in the metadata require a function
 function more(rest:Array<MetadataItem>>):haxe.macro.Function;

//else produce a final  
final notifier : MetadataAccess;
//or perhaps something else
function value(rest:Array<MetadataItem>):haxe.macro.Expr;
//path support
final path : {
  final into : MetadataAccess;
}
}
```

PR #113, opened 2023-10-05
0020-macro-sensitive-imports

0b1kn00b:
Thinking about macro errors and how to have less of them.

### DIFF
--- a/proposals/0020-macro-sensitive-imports.md
+++ b/proposals/0020-macro-sensitive-imports.md
@@ -1,0 +1,44 @@
+# ::enter feature name here::
+
+* Proposal: [HXP-0020](0020-macro-sensitive-imports.md)
+* Author: [0b1kn00b](https://github.com/ohmrun)
+
+## Introduction
+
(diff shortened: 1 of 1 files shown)


Simn:
I acknowledge the problem but I don't think this is the solution. Imports are just tools to simplify type paths, so whatever problem you can create with imports you can also create without them. As soon as you have one `macro function` in your file, there's a risk of running into the problems outlined here.

Simn:
I don't see how that argument (which I generally agree with) relates to the proposal. It's not like imports are inherited by the users of a library.

kLabz:
`MyModule.hx` vs `MyModule.macro.hx` is the safe way to separate macro and non macro code, without `#if macro` rendering code hard to read.

As Simn said, those imports you want to mark are not the only way to trigger problems so your proposal doesn't fix much.

nadako:
> MyModule.hx vs MyModule.macro.hx is the safe way to separate macro and non macro code

Can confirm, this kind of fixed all my issues with macro module dependencies.

PR #118, opened 2024-03-05
Improvements to Per-Project Haxelib Setups

EliteMasterEric:
Currently a draft PR as I hammer out the details on what the eventual solution for this specifically looks like (it would probably be a file with a name something like `haxelib.lock` in JSON format located at the root of the project folder), but would love feedback in the meantime.

### DIFF
--- a/proposals/0000-lock-file.md
+++ b/proposals/0000-lock-file.md
@@ -1,0 +1,40 @@
+# Improvements to Per-Project Haxelib Setups
+
+* Proposal: [HXP-0000](0000-lock-file.md)
+* Author: [EliteMasterEric](https://github.com/EliteMasterEric)
+
+## Introduction
+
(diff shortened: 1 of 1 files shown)


EliteMasterEric:
For reference, the `hmm.json` implements this like so::

```json
{
  "dependencies": [
    {
      "name": "library1",
      "type": "git",
      "dir": null,
      "ref": "2d83fa863ef0c1eace5f1cf67c3ac315d1a3a8a5",
      "url": "https://github.com/user/repo"
    },
    {
      "name": "library2",
      "type": "haxelib",
      "version": "1.1.0"
    }
}
```

Not sure at this point if we need anything else (notably, this allows for libraries to be defined as subdirectories of git repositories, which I believe `haxelib install all` currently does not provide for)

EliteMasterEric:
Closing because I'm dumb and didn't realize this exists:

https://github.com/HaxeFoundation/haxelib/pull/610

acarioni:
Another alternative is [lix](https://github.com/lix-pm/lix.client).

PR #122, opened 2024-07-16
Defer Keyword

DawDavis:
Proposal to add defer keyword, which would allow for similar functionality as in Go, obviating the need for `finally`, and for most types of destructor.

Rendered proposal [here](https://github.com/DawDavis/haxe-evolution/blob/defer/proposals/0000-defer.md)

### DIFF
--- a/proposals/0000-defer.md
+++ b/proposals/0000-defer.md
@@ -1,0 +1,189 @@
+# Add Defer
+
+* Proposal: [HXP-0000](0000-defer.md)
+* Author: [Dawson Davis](https://github.com/DawDavis)
+
+## Introduction
+
+`defer` is a keyword lifted directly from GoLang, which if implemented would allow for the automatic execution of an expression upon the exit of the current scope.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


Geokureli:
FWIW I find the example usage to be far, far less intuitive than a try catch.
```hx
mut.acquire();
defer mut.release();
someFallibleOperation();
```
The idea that line 3 happens before the conditionally executed line 2 goes against a fundamental understanding of how programming languages are expected to work.

IMO it would make a lot more sense to do something that preserves the order, like:
```hx
mut.acquire();
someFallibleOperation() defer mut.release();
```

I also think code using this is far more likely to cause human error, where new lines are added without noticeing that there is a defer further up in the method. In your next example, I see a multi-expression block of code that is flow controlled by the defer line.
```hx
exprs(); //any code preceding the defer

defer exprToDefer(args...);

body(); //any code after the defer, including a return
return x;
```
I think it should be absolutely mandatory that any conditionally executed block like this should be wrapped in curly brackets, maybe something like:
```hx
defer (exprToDefer(args...))
{
  body(); //any code after the defer, including a return
  return x;
}
```
Though, I still have issues:
1. This still has the ordering issues mentioned prior.
2. this doesn't seem much better than a try...catch.
3. it's not very clear on what's going on, in fact I really need clarification on what is returned here, or maybe I need a more concrete example

PR #123, opened 2024-11-12
[STANDARD LIBRARY BREAKING CHANGE] Enhance `sys.net`: add IP address types, support IPv6, update `Host` and friends

Frixuu:
> The networking-related APIs in the Haxe standard library currently differ significantly
> from those found in other languages such as C#, Java, or Rust.
> This proposal introduces changes to `sys.net` to align it with contemporary networking APIs
> while addressing several limitations in its current implementation.

[\<Rendered proposal\>](https://github.com/Frixuu/haxe-evolution/blob/sysnet/proposals/0000-ip-addresses-in-stdlib.md)

### DIFF
--- a/proposals/0000-ip-addresses-in-stdlib.md
+++ b/proposals/0000-ip-addresses-in-stdlib.md
@@ -1,0 +1,479 @@
+# Better support for IP addresses in the Haxe standard library
+
+* Proposal: [HXP-0000](0000-ip-addresses-in-stdlib.md)
+* Author: [Frixuu](https://github.com/Frixuu)
+
+## Introduction
+
+The networking-related APIs in the Haxe standard library currently differ significantly
+from those found in other languages such as C#, Java, or Rust.
+This proposal introduces changes to `sys.net` to align it with contemporary networking APIs
+while addressing several limitations in its current implementation.
+
+## Motivation
+
+### IPv6 support
+
+A significant limitation of the standard library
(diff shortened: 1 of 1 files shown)


Frixuu:
Request for voting.

Aidan63:
If there is a desire to add ipv6 support to the sys socket api then maybe it makes sense to have it utilise parts of the proposed asys api. Asys has support for ipv6 and has a v4 and v6 address enum and various tools very similar to what you've proposed.
Would seem rather silly to have duplicated functionality, maybe a `haxe.net.address` package for address classes / tools shared between sys and asys?

https://github.com/HaxeFoundation/haxe/blob/900dd50a4c840220cb9218a226d2736123557847/std/asys/native/net/Ip.hx

With asys looking to support ipv6 it might also be easier just to point people at that when they ask about ipv6 support instead of introducing breaking changes to the old sys api.

dimensionscape:
> With asys looking to support ipv6 it might also be easier just to point people at that when they ask about ipv6 support instead of introducing breaking changes to the old sys api.

To be fair, https://github.com/HaxeFoundation/haxe/pull/6142 issue was closed 5 years ago *in favor of the upcoming asys api*. But uh.. what is the status of asys these days?

@Frixuu I would be interested in working with someone on an intermediate library to shadow some of the api and fill in the gaps to provide some utility here in the mean time (and provide compatibility with these features for Haxe 4.x.x).

PR #125, opened 2024-12-20
abstract + enum + partial order = automatic constuctors 

0b1kn00b:
Using the type system to create automated constructors without substantially impacting the AST is possible, includes working code which demonstrates the approach.


RblSb:
I don't get what you want to add in this proposal. Can you provide inline examples with and without your feature to show benefits and explain what it does?

0b1kn00b:
If you order the AST by shortlex ordering, the compiler should be able to derive the grand majority of constructors with that information alone, I don't know if you've seen the macro di implementation but it's not super complex.

